PR #9, opened 2019-12-15
Fix import typo

plasticviking:
A typo in a previous PR broke the frontend. This fixes it.


amichard:
The change in capitalization was actually intentional. The fix here is to rename it to something else then back to App.js

PR #21, opened 2020-01-09
Initial development for django back-end with integration to the front…

amichard:
for ZEVA-47, ZEVA-28

Changelog:
- Django scaffolding
- Integrated to front-end
- Initial migration
- Organization details should load most data from the backend


backend/zeva/urls.py
```diff
@@ -0,0 +1,22 @@
...
+    2. Add a URL to urlpatterns:  path('', Home.as_view(), name='home')
+Including another URLconf
+    1. Import the include() function: from django.urls import include, path
+    2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
+"""
+from django.contrib import admin
+from django.urls import include, path
+
+urlpatterns = [
+    path('admin/', admin.site.urls),
```
plasticviking: Are we going to use the Django admin functionality?

plasticviking:
Oh, and `db.sqlite3` too

backend/zeva/urls.py
amichard: We can disable this for now. I've been wondering if this will be useful somehow. But for now, I'll just disable it.

PR #23, opened 2020-01-09
isGovernment attribute for checking whether user is IDIR or BCEID

amichard:
follow-up for ZEVA-28

Changelog:
- Added fixture for adding Government as an organization
- Added isGovernment column for organizations
- Added isGovernment property for User 
- Move token for authorization as a configuration
- Updated navbar and other sections to read from the call to the user


kuanfandevops:
Close this pr as the files are included in pr#24

PR #30, opened 2020-01-13
Vehicle Information

plasticviking:
# Changelog
 - Create models for vehicle information and related entities
 - Preliminary viewset and serializer, list view in frontend (basic for now)
 - Remove last vestiges of grpc code
 - Create operational scripts to enable local experimentation (works in `skaffold`/`kubernetes` too, but you need to get a shell into the django pod to run it -- it should be persistent).


backend/api/models/mixins/Named.py
```diff
@@ -0,0 +1,24 @@
+from django.db import models
+
+from api.managers.uniquely_named import UniquelyNamedManager
+
+
+class Named(models.Model):
+    name = models.CharField(blank=False,
```
klefler: Could we call this "description"? We would typically want to shy away from using keywords as column names.
If so, does this impact the class name? Is it possible to default a column column as well.

backend/api/models/mixins/Named.py
```diff
@@ -0,0 +1,24 @@
...
+                            null=False,
+                            max_length=250)
+
+    class Meta:
+        abstract = True
+
+
+class UniquelyNamed(models.Model):
+    name = models.CharField(blank=False,
```
klefler: Same as above... Could we call this "description" and ensure a column comment is added?

backend/api/models/credit_value.py
```diff
@@ -0,0 +1,18 @@
...
+from .mixins.Named import UniquelyNamed
+from .organization_address import OrganizationAddress
+from .user_profile import UserProfile
+from ..managers.organization import OrganizationManager
+
+
+class CreditValue(Auditable):
+
+    class Meta:
+        db_table = 'credits'
```
klefler: Table name should be singular and there should be a db_comment to describe the table. Would it make sense to call this table "something"_credit? or would "credit" suffice?

backend/api/models/credit_value.py
```diff
@@ -0,0 +1,18 @@
...
+
+    a = models.DecimalField(null=True, blank=False, decimal_places=3, max_digits=5)
```
klefler: Is this a column called 'a'? If so, is there a more descriptive name we could use? Also need to add a column description.

backend/api/models/credit_value.py
```diff
@@ -0,0 +1,18 @@
...
+    b = models.DecimalField(null=True, blank=False, decimal_places=3, max_digits=5)
```
klefler: Same as above.

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+from .credit_value import CreditValue
+from .mixins.Named import UniquelyNamed, Named
+from .mixins.effective_dates import EffectiveDates
+
+
+class ModelYear(Auditable, EffectiveDates, UniquelyNamed):
+
+    class Meta:
+        db_table = 'model_year'
+
```
klefler: Please add a db_comment with a few examples of the values that would stored in this table.

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+
+class Type(Auditable, Named):
+
+    class Meta:
+        db_table = 'type'
```
klefler: Recommending we call this table vehicle_type as "type" is not descriptive and could only be used once.

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+
+    db_table_comment = 'Set of all vehicle Types'
```
klefler: Please add a couple of examples of the values we'd see in this table.

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class Make(Auditable, Named):
+
+    class Meta:
+        db_table = 'make'
```
klefler: Recommending we call this table vehicle_make.

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+
+    db_table_comment = 'Set of all vehicle Makes'
```
klefler: Could you please add a few examples of lookup values in this table?

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class Model(Auditable, Named):
+
+    class Meta:
+        db_table = 'model'
```
klefler: Recommend that we name this table vehicle_model_type or model_type.

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+
+    make = models.ForeignKey(
+        Make,
+        related_name='valid_models',
+        on_delete=models.DO_NOTHING,
+        null=False
+    )
+
+    db_table_comment = 'Set of all vehicle Models'
```
klefler: Could you please add a few examples to the comment?

klefler: Another thought... Will there be a need to create a lookup that ties together make/model/trim? Like a validated list? Does it make sense to store these values in separate tables? How could we validate that the correct model/trim was selected for a make?

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class Trim(Auditable, Named):
+
+    class Meta:
+        db_table = 'trim'
```
klefler: Recommend naming this table vehicle_trim_type or trim_type.

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+from .mixins.effective_dates import EffectiveDates
+
+
+class ModelYear(Auditable, EffectiveDates, UniquelyNamed):
+
+    class Meta:
+        db_table = 'model_year'
+
+
+class Type(Auditable, Named):
```
klefler: Should this have the EffectiveDates?

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+
+    class Meta:
+        db_table = 'type'
+
+    db_table_comment = 'Set of all vehicle Types'
+
+
+class Make(Auditable, Named):
```
klefler: Should this have the EffectiveDates?

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+
+    class Meta:
+        db_table = 'make'
+
+    db_table_comment = 'Set of all vehicle Makes'
+
+
+class Model(Auditable, Named):
```
klefler: Should this have the EffectiveDates?

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+        Make,
+        related_name='valid_models',
+        on_delete=models.DO_NOTHING,
+        null=False
+    )
+
+    db_table_comment = 'Set of all vehicle Models'
+
+
+class Trim(Auditable, Named):
```
klefler: Should this have the EffectiveDates?

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+        db_table = 'trim'
+
+    model = models.ForeignKey(
+        Model,
+        related_name='valid_trims',
+        on_delete=models.DO_NOTHING,
+        null=False
+    )
+
+    db_table_comment = 'Set of all vehicle Trims'
```
klefler: Could you please add a few examples of trim values we'd see in this table?

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+        on_delete=models.PROTECT
+    )
+
+    type = models.ForeignKey(
+        Type,
+        related_name=None,
+        on_delete=models.PROTECT
+    )
+
+    range = models.IntegerField(
```
klefler: Recommend naming this column vehicle_range? I think 'range' may be a keyword.

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+        db_comment='Range in km'
```
klefler: Maybe just add 'vehicle range in km'

backend/api/models/vehicle.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+
+    model_year = models.ForeignKey(
+        ModelYear,
+        related_name=None,
+        on_delete=models.PROTECT,
+        null=False
+    )
+
+    validated = models.BooleanField(default=False,
```
klefler: We should have one consistent way to name boolean columns... I think we have used "is_" or "has_".... so could this column be named 'is_validated'?

backend/api/models/organization.py
```diff
@@ -9,11 +9,6 @@
 
 
 class Organization(Auditable):
-    """
-    Contains a list of all of the recognized Part 3 fuel suppliers, both
-    past and present, as well as an entry for the government which is also
-    considered an organization.
-    """
     name = models.CharField(
```
klefler: Recommend naming this column 'organization_name' as 'name' is a keyword.

backend/api/models/organization.py
```diff
@@ -9,11 +9,6 @@
 
 
 class Organization(Auditable):
```
klefler: Should we consider adding the EffectiveDates? Or some way to determine if a supplier is "active" or not?

PR #31, opened 2020-01-16
Submit vehicle info backend

amichard:
ZEVA-30

Changelog:
- 1 class 1 model
- Frontend BaseController for common functionality like handle input change
- Fixed token renewal from the front-end
- Connected front-end to back-end, but since the fields are just regular input fields
- General code cleanup
- Please note the rename of the capitalization change for named.py
- Updated DB comments


backend/api/models/vehicle.py
```diff
@@ -1,85 +1,29 @@
 from django.db import models
 
 from auditable.models import Auditable
-from .credit_value import CreditValue
```
plasticviking: I don't have any issue with moving these into separate files, though I also don't really understand the benefit. The code is pretty tightly-related.

backend/api/serializers/vehicle.py
```diff
@@ -1,38 +1,82 @@
...
+        )
+
 
+class VehicleSerializer(serializers.ModelSerializer):
     credit_value = CreditValueSerializer()
-    type = serializers.SlugRelatedField(slug_field='name', read_only=True)
-    make = serializers.SlugRelatedField(slug_field='name', read_only=True)
-    model = serializers.SlugRelatedField(slug_field='name', read_only=True)
-    trim = serializers.SlugRelatedField(slug_field='name', read_only=True)
-    model_year = ModelYearSerializer()
```
plasticviking: I think `ModelYear` is going to be something more general -- like a compliance period. I'm not sure it makes sense to prepend `Vehicle` to its name.

frontend/src/vehicles/components/VehicleListTable.js~master
```diff
@@ -0,0 +1,108 @@
+/*
```
plasticviking: Leftover merge file?

klefler:
Is this the latest PR in the works? There were some changes that @amichard and I were looking at yesterday afternoon (after this PR was submitted)... and I don't see them here.

backend/api/models/vehicle.py
amichard: It's generally just for organization. Subscribing to the one class per model rule.

amichard:
> Is this the latest PR in the works? There were some changes that @amichard and I were looking at yesterday afternoon (after this PR was submitted)... and I don't see them here.

Sorry, Kristin. I haven't committed them yet. I'll commit them now.

backend/api/serializers/vehicle.py
amichard: Sounds good. I'll change it to model_year.py

frontend/src/vehicles/components/VehicleListTable.js~master
amichard: It was. Sorry about that. Removing the file.

PR #34, opened 2020-01-20
Review eligibility requests

plasticviking:
# Changelog
 - Remove `validated` in favour of state enum [NEW, DRAFT, SUBMITTED, VALIDATED, REJECTED]
 - Create history table and basic authorization checks
 - `state_change` detail route and serializer
 - maintain history of state changes (basic UI exposed for now)
 - export possible status changes as `actions` field of returned `vehicle` dict


amichard:
Tagging @klefler 

Also, can we run eslint here?

backend/api/models/vehicle.py
```diff
@@ -58,8 +58,16 @@ class Meta:
     db_table_comment = 'Set of all vehicle Trims'
 
 
-class Vehicle(Auditable):
+@unique
+class VehicleDefinitionStates(Enum):
```
klefler: Question - should these values be stored in a status table? Not doing so could make the data hard to manager over time.

backend/api/models/vehicle.py
```diff
@@ -95,9 +103,12 @@ class Vehicle(Auditable):
         null=False
     )
 
-    validated = models.BooleanField(default=False,
-                                    db_comment='Whether this vehicle description has been validated'
-                                               ' by the regulator')
+    state = EnumField(VehicleDefinitionStates,
```
klefler: related to question above, would prefer this column name be changed, something like 'validation_status' or 'vehicle_status' or ... if a lookup table is done, name should match.

backend/api/models/vehicle.py
```diff
@@ -106,3 +117,30 @@ class Meta:
         unique_together = [['make', 'model', 'trim', 'model_year']]
 
     db_table_comment = "List of credit-generating vehicle definitions"
+
+
+class VehicleChangeHistory(Auditable):
```
klefler: Depending on whether or not there's a decision to change the roles and/or vehicle states model, we might want to re-look at this and discuss various patterns for storing history records.

backend/api/models/user_profile.py
```diff
@@ -61,6 +63,12 @@ class UserProfile(Auditable):
         db_comment="Username that we can connect the user to Keycloak."
     )
 
+    # Unpersisted (supplied by Keycloak authentication filter)
+    roles = []
```
klefler: curious about this column... would it contain one role or many? If many, would it be better to normalise the data?

Should consider changing the column name, not plural, and possibly more descriptive... 'user_role'.

Needs db_comment please.

plasticviking: This isn't actually a DB column, @klefler, just plain Python. It's only used at runtime to store the value loaded from Keycloak and is never saved.

backend/api/models/vehicle.py
plasticviking: I think that's something we struggled with in TFRS. Application logic (especially state transitions) are going to depend on these values and I think it makes sense to leave them in code since there is not realistically a case where you'd update the database and not the code (or if you did update the db, especially to add new values, it could break the code logic and would need to be unit tested).

backend/api/models/user_profile.py
klefler: Thanks, apologies for the mistake.

backend/api/models/vehicle.py
klefler: Fair enough. In this case, it sounds like the statuses will be fairly stable as well, so the risk would be low... thanks for the clarification.

PR #36, opened 2020-01-23
List approved vehicles

amichard:
ZEVA-21

Changelog:
- Added relationship table between vehicle_make and organization
- Minor code cleanup


backend/api/migrations/0014_auto_20200123_1634.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0013_vehiclechangehistory_at'),
+    ]
+
+    operations = [
+        migrations.AddField(
```
plasticviking: Since we've now got the state flag, I think it makes more sense to have `is_validated` be a property (that just checks if state is `VehicleDefinitionStates.VALIDATED`. Creating an additional DB field opens us up to having to check a bunch of edge cases (like state DRAFT but validated true).

amichard: I just noticed this. I'll have to check why the migration is pulling that back in. Might've been a merge conflict.

amichard: Thanks for catching this, Rob. is_validated should be gone now.

PR #48, opened 2020-02-04
Validation status

amichard:
ZEVA-68

Changelog:
- Pylint/Pycodestyle fixes
- Removed unused models
- Removed old test fixtures
- Changed state into validation_statuses
- Generated new migration
- Updated history table


backend/api/models/vehicle_change_history.py
```diff
@@ -0,0 +1,76 @@
...
+from django.contrib.postgres.fields import ArrayField
+from django.db import models
+
+from auditable.models import Commentable
+from enumfields import EnumField
+from .vehicle_statuses import VehicleDefinitionStatuses
+
+
+class VehicleChangeHistory(Commentable):
+    make = models.ForeignKey(
```
plasticviking: If the goal is to track the history of changes, shouldn't we be saving the relevant chunks of foreign data directly in our table? Otherwise they're not immutable, unless the logic creates duplicates of foreign entities when a history item is created.

It's not clear to me from reading the code what the intent is here.

amichard: Yes, it would be a problem if someone decides to modify the data the foreign key is referring to, but at that point the question would be why are we allowing edits to the data it's referencing to instead of adding new values if they mean something else. 

This history table is pretty much the same as the vehicle table. The only difference is that it doesn't have an updated_timestamp and updated_user as it will always create a new record (no updates allowed in the history table). For every change that happens to the vehicle table, the history table is supposed to keep track of it. ie Every time we hit create/update it creates a record of the values in the history table. That way we can always go back on what the values were at the specific time.

amichard: After discussing with Kristin, we decided that we'll change this so that it saves the foreign data directly.

backend/api/models/vehicle_change_history.py
```diff
@@ -0,0 +1,76 @@
...
+                        statuses=[c.name for c in VehicleDefinitionStatuses]
+                   )
+    )
+    vehicle = models.ForeignKey(
+        'Vehicle',
+        null=True,
+        on_delete=models.CASCADE,
+        related_name='history'
+    )
+    in_roles = ArrayField(
```
klefler: Suggest renaming column to 'user_role'.

**Comment:** We should be aware that this column would be storing a "one-to-many" type string. This is fine if the column is used for just information/display purposes, but could pose potential issues if any processing/reporting/filtering is required... If there is no requirement to use this column for filtering, etc. then there's low risk, but, if that changes we'd likely want to re-address how this data is stored.

backend/api/models/vehicle_change_history.py
```diff
@@ -0,0 +1,76 @@
...
+        models.CharField(
+            max_length=50,
+            blank=False,
+            null=True
+        ),
+        db_comment='The roles the actor had at the moment the status changed'
```
klefler: We might want to update the comment to make it really clear what this column's purpose is.

'The role (or roles) assigned to the user at the time they made a change to a vehicle record. Role values are generated from Keycloak and are stored in a comma-delimited string.'

Note: not sure if these are stored as comma-delimited or not, but it might be helpful to state this.

backend/api/models/vehicle_change_history.py
amichard: I'll remove the restriction and just store it as a regular integer

amichard: To clarify, we store the ID of the table still as we really shouldn't be allowing users to just modify data in our foreign key tables. They should be adding new ones and expiring the old one, instead (unless they're just minor changes such as spelling). I do get the idea behind saving the data directly into the history table, though. We can discuss it further, if you'd like.

backend/api/models/vehicle_change_history.py
amichard: Comment updated

backend/api/models/vehicle_change_history.py
amichard: column has been renamed to user_role

PR #51, opened 2020-02-05
Disabled audience detection as it's causing issues in the dev server

amichard:
Disabling the verify_aud as it's causing issues in dev server


plasticviking:
This is an important security measure. It should not be disabled, even for debugging. If the audience value is not correct, we need to fix it.

amichard:
I'd like to know why this check is important. What does the audience check actually do?

PR #55, opened 2020-02-06
Followup: Code cleanup for login fix

plasticviking:
# Changelog
 - Move proptypes for common things to a utility file to avoid boilerplate
 - linting
 - use `implicit` oauth2 flow


backend/api/fixtures/development/0000_add_sample_users.py
```diff
@@ -0,0 +1,39 @@
+from django.db import transaction
```
amichard: Is there a reason this is in development and not in the test folder?

PR #64, opened 2020-02-10
Fixed load_ops_data tests

amichard:
I was encountering errors with the system defaults being an unknown argument.

So I'm just specifying the arguments now. If this shouldn't be showing errors, then I'll just close the PR


amichard:
Closing this. It was my python version causing the problems

PR #80, opened 2020-02-19
zeva 84- user profile frontend

emi-hi:
front end for user profile
-user form fills with user data from api
-changes to form updates state for user info and roles
-very simple validation to just check if field is empty 
-no submit functionality yet


kuanfandevops:
@emi-hi looks like there are some conflicts in router.js

kuanfandevops:
close it and merge 87

PR #82, opened 2020-02-24
Sales data model (preliminary)

plasticviking:


backend/api/models/record_of_sale.py
```diff
@@ -0,0 +1,78 @@
...
+    vehicle = models.ForeignKey(
+        'Vehicle',
+        related_name='sales',
+        on_delete=models.CASCADE
+    )
+
+    sale_id = models.CharField(blank=True,
+                               null=True,
+                               max_length=255,
+                               db_comment="An (optional) vehicle-supplier-provided correlation ID")
```
klefler: Might need to update this comment... would this be a sales number like an invoice number from the supplier? Or a unique number from the supplier's database?

backend/api/models/record_of_sale.py
```diff
@@ -0,0 +1,78 @@
...
+
+    vin = models.CharField(blank=True,
```
klefler: Could we rename this column to vehicle_identification_number ... or update the comment to describe VIN (see comment below)? Also wondering why it's optional. We may want to consider limiting the max length to something smaller than 255... I've read that the max length of a VIN is 17 characters. Would it make sense to limit the length to that?

backend/api/models/record_of_sale.py
```diff
@@ -0,0 +1,78 @@
...
+                           null=True,
+                           max_length=255,
+                           db_comment="An (optional) vehicle-supplier-provided correlation ID")
```
klefler: Should have a more descriptive comment... for example (from ICBC):

"A vehicle identification number is a series of letters and numbers that vehicle manufacturers assign to each vehicle they make. The VIN is a unique, 17-digit number used to identify a vehicle, including its origin, year, make and model, and production number."

backend/api/models/record_of_sale.py
```diff
@@ -0,0 +1,78 @@
...
+
+    vin_validation_status = EnumField(
```
klefler: Would prefer this column be driven by a lookup table... the concern being management of the code/values over time (if they are "no longer valid").

backend/api/models/record_of_sale.py
```diff
@@ -0,0 +1,78 @@
...
+                   "{statuses}".format(
+            statuses=[c.name for c in VINStatuses]
+        )
+    )
+
+    sale_date = models.DateField(blank=True,
+                                 null=True,
+                                 db_comment="The calendar date the vehicle was placed with an end customer")
+
+    submission_date = models.DateField(
```
klefler: Is this date automatically created when the supplier submits their sales records? And does this date ever change? For example; if the supplier needs to make a change to a record that was previously entered?

backend/api/models/record_of_sale.py
```diff
@@ -0,0 +1,78 @@
...
+        blank=True,
+        null=True,
+        auto_now=True,
+        db_comment="The calendar date the submission was entered in ZEVA"
+    )
+
```
klefler: For validation_status... the validation is on a single record submitted for credit rather than a submission of many records? Will it be difficult to manage statuses of 1000's of records being submitted? Just clarifying as I was trying to see if there is a need for a "submission" which contains many VINs... rather than each VIN "is" a submission.

backend/api/models/record_of_sale.py
```diff
@@ -0,0 +1,78 @@
...
+    validation_status = EnumField(
```
klefler: As mentioned above, would prefer this be a lookup table.

backend/api/models/record_of_sale.py
```diff
@@ -0,0 +1,78 @@
...
+    # transaction = models.ForeignKey(
+    #         'CreditTransaction',
+    #         related_name='sale',
+    #         on_delete=models.PROTECT,
+    #         null=True,
+    #     )
+
+    class Meta:
+        db_table = "record_of_sale"
+        unique_together = [['organization', 'sale_id']]
```
klefler: If the sale_id is nullable, not sure this would work as a unique constraint.

Wondering, is the VIN unique? Or can a supplier re-submit a VIN if it isn't validated when initially submitted?

klefler:
Hi @plasticviking - Thanks for sending the new model! I hope my comments/questions make sense. If not, please let me know. I was hoping to find the latest screen designs (they may answer some of my questions)... but Al was off today so I had to wing it a bit.

backend/api/models/record_of_sale.py
plasticviking: It could be anything like that. I thought it might be useful for them to store a correlation ID to their own systems (such as an invoice number). I don't suggest we assign any particular semantics to it, but use it as a search/sort field visible to vehicle suppliers. Does that make sense?

backend/api/models/record_of_sale.py
plasticviking: Yes, I will update this one. Good catch.

backend/api/models/record_of_sale.py
plasticviking: This should be the invoice date (the date the record is submitted is entered separately, and calculated automatically). Whether or not it can change if entered erroneously is a business decision -- I can see having the desire to modify this value when the record is in draft state.

backend/api/models/record_of_sale.py
plasticviking: Al and I had a discussion about this. I'm not sure what storing the submission separately would gain us, and we can always group them by submission timestamp/submitting user if we need to aggregate submissions. I don't think there are any properties of a submission we'd need to track (but if there are, it's simple enough to add later)

backend/api/models/record_of_sale.py
plasticviking: My concern here is my usual one -- any kind of "status" table isn't really a lookup table, it can drive application logic (the actions available to each interested party in the workflow change as it changes state). There's no way to cleanly separate or sensibly expire these values that makes sense outside of the context of the code.

backend/api/models/record_of_sale.py
plasticviking: I think we have to allow non-unique VINs, and catch them with validation logic (a "duplicate-VIN" status).

plasticviking: > If the sale_id is nullable, not sure this would work as a unique constraint.

Django unique constraints are not 1:1 with database-level unique constraints. They explicitly permit null values (and null values are ignored when doing uniqueness checks).

backend/api/models/record_of_sale.py
klefler: Sure, yep that makes sense. I don't want to get too finicky over column naming, but I'm wondering if we can find a different name. The suffix of "id" has some common usage within NRM databases, such as to denote a relationship to another table... or a surrogate key id value. Could we possibly change the name of this column to something like '_sale_reference_number'_? Or if it's too much work to do that, if the column comment should be updated to something like "An (optional) id/reference number provided by the vehicle supplier, used as a reference to their own sales systems; such as an invoice number."

backend/api/models/record_of_sale.py
klefler: Ok fair enough. Makes sense. I think I might want some clarity on the submission_date column above...

backend/api/models/record_of_sale.py
klefler: I think I many be confused about the difference between sale_date and submission_date.

I assumed that the **sale_date** column is the supplier's invoice date... and that value would entered by the supplier in their submission. Is this correct?

I interpreted the **submission_date** as the date the record of sale was added to ZEVA by the supplier (as mentioned in the db comment). Kind of what you described above... as the automatically added value. Is that correct?

The reason I'm asking for clarity on the dates (and whether or not the submission date can change) is that I'm trying to understand the concept of "submission" as well as any potential keys that could make a record_of_sale record unique. And I'm trying to understand the workflow and how the data may change based on potential scenarios... for example; is something ever re-submitted (and if so, does the submission date for the record stay the same? Does the submission date get updated for an existing record? Or is it a new submission record?

backend/api/models/record_of_sale.py
klefler: Sorry for dragging this one out... Perhaps I am misunderstanding everything that's involved with the workflows, etc. From how I see it, either way, there would be major impacts to changing these values (be it in a database table or in the code)... My main concern is how these values are managed over time and how the data maps to values that are no longer be valid. How is history maintained? Does it need to be? I'm fine with the design as-is if you don't think there is any risk here.

plasticviking: I don't think there's any risk here.

I understand the desire to keep things separable when they are, but these values make no sense outside of the context of the code version that goes with them.

If they become irrelevant at some point in the future, one of two things will happen:

1) We just stop assigning the deprecated statuses in some future version of the code. The old records can stay, but no new ones will be created with that status.
2) We create a migration that does something interesting with rows having the deprecated status (maybe archiving them or moving them to some new status) and actually delete the old status.

In either case, there's a code change to support the change in status -- it's an action that doesn't make sense outside of the code+data system.

To expand on why I don't like lookup tables for things like this:
Our code is full of logic like this (pseuocode):

```
if entity->status is NEW and current_user has role 'SUBMIT_RECORD' then:
 set entity->status to SUBMITTED
```

If we use hardcoded values for these statuses, we can benefit from language-level checking within our tools (for example, if we typo the status on line 2 of this example as `SUMBITTED`, our tests, IDES, linting, etc, can catch it before it ever goes near a database). The code can be understood without reference to external entities.

If we store it in the database, we effectively have to guard our code with `SELECT`s to achieve the same result. Like:

```
new_status = orm.statuses.get(name like 'NEW').exists ? orm.statuses.get(name like 'NEW') otherwise NULL
submitted_status = orm.statuses.get(name like 'SUBMITTED').exists ? orm.statuses.get(name like 'NEW') otherwise NULL

if new_status is  NULL or submitted_status is not  then:
 raise Error('missing status in DB')

if entity->status is new_status and current_user has role 'SUBMIT_RECORD' then:
 set entity->status to submitted_status
```
It results in a lot of extra boilerplate, code that is less safe, and doesn't change the fact that the code and data are linked.

backend/api/models/record_of_sale.py
plasticviking: Your analysis of the semantics of the date values matches my intention.

I imagine we would follow our typical pattern of having basically immutable rows (or immutable once the state changes to SUBMITTED). I don't think the submission date can ever change -- I would expect it to be the date on which the status changes to `SUBMITTED`, (so that DRAFT records, visible only to fuel suppliers, are not yet counted as having a submission date). I expect we'll replace the `auto_now` parameter with application logic that sets the date when the status changes.

I have not heard any requirement to track the history of individual items, but it is conceivable we might be asked to do this. Would you prefer to see a more robust history-tracking mechanism here, similar to the one for the vehicle table, instead of just the date submitted?

klefler: Thanks Rob, that explanation helps. I don't think we need to add anything additional here. If down the road, James wants history, we can address it then. Just having the clarity of "submission date" helped me. If the submission date is set upon SUBMITTED status, maybe it would help to have the column comment reflect that. For example; _The calendar date a supplier submitted a record of ZEV sale to the ministry_. (or something like that).

backend/api/models/record_of_sale.py
klefler: Thanks! Great explanation. Definitely helps.

kuanfandevops:
switch the target branch from master to release-sprint-6

PR #97, opened 2020-03-02
Add run preconditions, possibly python 3.6 fix

plasticviking:
# Changelog
 - Fixes for operational data scripts
 - Runs now actually idempotent
 - Fixed a possible Python 3.6 compatibility issue

### DIFF
--- a/backend/api/fixtures/operational/0001_add_vehicle_makes.py
+++ b/backend/api/fixtures/operational/0001_add_vehicle_makes.py
@@ -12,22 +12,28 @@ class AddVehicleMakes(OperationalDataScript):
     is_revertable = False
     comment = 'Adds the vehicle makes found in the NRCAN document'
 
+    list_of_makes = [
+        "Aston Martin", "Audi", "BMW", "Chevrolet", "Chrysler",
+        "Ferrari", "Fiat", "Ford", "GMC", "Honda", "Hyundai", "Isuzu",
+        "Jaguar", "Karma", "Kia", "Mazda", "Mercedes-Benz", "Mini",
+        "Mitsubishi", "Nissan", "Porsche", "Smart EQ", "Subaru", "Suzuki",
+        "Tesla", "Toyota", "Volkswagen", "Volvo"
+    ]
+
     def check_run_preconditions(self):
+        for make_name in self.list_of_makes:
+            if Make.objects.filter(name=make_name).exists():
+                return False
+
         return True
 
     @transaction.atomic
     def run(self):
-        list_of_makes = [
-            "Aston Martin", "Audi", "BMW", "Chevrolet", "Chrysler",
-            "Ferrari", "Fiat", "Ford", "GMC", "Honda", "Hyundai", "Isuzu",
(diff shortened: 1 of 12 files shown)


backend/api/fixtures/operational/0001_add_vehicle_makes.py
```diff
@@ -12,22 +12,28 @@ class AddVehicleMakes(OperationalDataScript):
...
+    list_of_makes = [
+        "Aston Martin", "Audi", "BMW", "Chevrolet", "Chrysler",
+        "Ferrari", "Fiat", "Ford", "GMC", "Honda", "Hyundai", "Isuzu",
+        "Jaguar", "Karma", "Kia", "Mazda", "Mercedes-Benz", "Mini",
+        "Mitsubishi", "Nissan", "Porsche", "Smart EQ", "Subaru", "Suzuki",
+        "Tesla", "Toyota", "Volkswagen", "Volvo"
+    ]
+
     def check_run_preconditions(self):
+        for make_name in self.list_of_makes:
```
amichard: Is this supposed to fail the script if it sees a make that's already in the database?

The script is supposed to run even when the make already exists and should just skip it, instead of failing the whole script

backend/api/fixtures/operational/0005_add_organizations.py
```diff
@@ -13,73 +13,79 @@ class AddOrganizations(OperationalDataScript):
...
+        "name": "Volkswagen Group Canada Inc.",
+        "makes": ["Volkswagen"]
+    }, {
+        "name": "Volvo Car Canada Ltd.",
+        "makes": ["Volvo"]
+    }]
+
+
     def check_run_preconditions(self):
+        for org in self.list_of_organizations:
```
amichard: Same as above. I don't think the whole script should fail, when it can just add the ones that are missing.

backend/api/fixtures/test/0001_add_plugin_hybrid_vehicles.py
```diff
@@ -17,162 +17,166 @@ class AddPluginHybridVehicles(OperationalDataScript):
...
+    }, {
+        "make": "Volvo",
+        "model": "XC90 T8 AWD",
+        "fuel_type": "BZ",
+        "range": "27",
+        "class_code": "UL"
+    }]
+
     def check_run_preconditions(self):
+        for veh in self.list_of_vehicles:
```
amichard: why veh?

backend/api/fixtures/operational/0005_add_organizations.py
plasticviking: Addressed in push

backend/api/fixtures/operational/0001_add_vehicle_makes.py
plasticviking: Addressed in push

backend/api/fixtures/test/0001_add_plugin_hybrid_vehicles.py
plasticviking: bits are expensive

PR #99, opened 2020-03-03
Sales Frontend, Template Upload

plasticviking:
# Changelog
 - Create sales excel templates (via web or command line)
 - Upload those same templates (Still needs some polish)
 - View the sales data in the frontend

JIRA Cards 66, 76, 82, 83 (still some polish to be added though)


backend/api/management/commands/generate_sales_template.py
```diff
@@ -0,0 +1,45 @@
...
+        helptext = ('Generate Excel sales templates for a given Organization (omit the name for a list)')
+
+        parser.description = helptext
+
+    def handle(self, *args, **options):
+        org_name = options['organization_name']
+        file_name = options['file_name']
+        if org_name is None:
+            for o in Organization.objects.filter(is_government=False):
+                print(o.name)
```
amichard: Why does this print the list of organizations when the argument is not used?

backend/api/services/sales_spreadsheet.py
```diff
@@ -0,0 +1,226 @@
...
+        make = ma.vehicle_make
+        logger.info('{org} supplies {name}'.format(org=organization.name,
+                                                   name=make.name))
+
+        vehicles = Vehicle.objects.filter(make=make)
+        for veh in vehicles:
+            logger.info('{make} has model {model}'.format(make=make.name,
+                                                          model=veh.model_name))
+
+            sheet_name = '{} - {}'.format(make.name, veh.model_name)
```
amichard: I'm running into a duplicate worksheet name. Adding model year id still gave me the error, but when I added the vehicle id, that seem to make the sheet name unique enough that it doesn't complain

backend/api/management/commands/generate_sales_template.py
plasticviking: So you can see what name to supply as an argument, since it has to be character-perfect. It's just a convenience.

backend/api/services/sales_spreadsheet.py
plasticviking: Ah, I guess we need to differentiate when two vehicles have the same make and model. I don't have any in my database that meet that criteria. I'll fix that.

PR #100, opened 2020-03-03
Update for testing ops scripts

kuanfandevops:


frontend/src/app/components/Footer.js
```diff
@@ -51,7 +51,7 @@ const Footer = () => (
         activeClassName="active"
         to="/"
       >
-        Contact Us
+        Contact Us TTT
```
amichard: What does TTT mean?

PR #105, opened 2020-03-06
Sales Submissions Data Model

plasticviking:
# Changelog
 - Data model for submissions (batches of sales submissions -- a file upload)
 - Appropriate migrations to preserve old records


backend/api/models/sales_submission.py
```diff
@@ -0,0 +1,39 @@
+from django.db import models
+from enumfields import EnumField
+
+from api.models.mixins.named import UniquelyNamed
+from api.models.sales_submission_statuses import SalesSubmissionStatuses
+from auditable.models import Auditable
+
+
+class SalesSubmission(Auditable, UniquelyNamed):
```
klefler: Hi Rob. Just one question re: name column. Could we create a column directly instead? Maybe called submission_name, or similar? Not sure the generic nature of the UniquelyNamed column provides enough info. Specifically the column comment. What is this name column used for?

plasticviking: As I think about it, I think it makes more sense to have a composite key, composed of submission date, organization, and a sequential number for an organization/date (A, B, C...) if there's more than one. I think that captures everything we might need to generate the submission IDs.

klefler: For the composite key, is that to help with creating a "reference number" for the submission?
Is this the key you would pass to the record_of_sale table? Is it going to be passed as one column or 3 columns?

And is this in lieu of having a "name" column then?

Sorry for all of the questions. Just not totally clear. Thanks.

PR #109, opened 2020-03-10
Sales submissions

plasticviking:
# Changelog
 - Added a submission table for tracking batch sales uploads
 - Starting to use React Table pivoting for displaying aggregate data in the lists of vehicles
 - Fixed issue when importing numeric VIN

JIRA: ZEVA 71, 74, 75


backend/api/models/sales_submission.py
```diff
@@ -0,0 +1,84 @@
...
+        null=True,
+        auto_now_add=True,
+        db_comment="The calendar date the submission was created in ZEVA"
+    )
+
+    submission_sequence = models.CharField(
+        blank=False,
+        null=True,
+        max_length=3,
+        db_comment=''
```
klefler: The comment is missing. Please add when you have a chance.

backend/api/models/sales_submission.py
```diff
@@ -0,0 +1,84 @@
...
+        chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+
+        if count == 0:
+            return None
+
+        x = count
+
+        seq = ''
+
+        while True:
```
amichard: Do the characters mean anything specifically? I guess I'm checking on why we can't use sequential numbers, instead.

PR #114, opened 2020-03-12
Credits Data Model

plasticviking:
# Changelog
 - Submitting for feedback my take on a suitable data model for credits, credit classes, organization balances
 - fixed a frontend compilation-blocking issue introduced in another PR


backend/api/models/credit_class.py
```diff
@@ -0,0 +1,19 @@
+from django.db import models
+
+from api.models.mixins.effective_dates import EffectiveDates
+
+class CreditClass(EffectiveDates):
+    class Meta:
+        db_table = "credit_class"
```
klefler: Could this be renamed credit_class_code, to keep consistency with other lookups?

backend/api/models/credit_class.py
```diff
@@ -0,0 +1,19 @@
+from django.db import models
+
+from api.models.mixins.effective_dates import EffectiveDates
+
+class CreditClass(EffectiveDates):
```
klefler: The other lookup tables have Auditable and Description... would recommend adding Auditable, but Description could be up for discussion... although it would be good to have a short description with the code value.

backend/api/models/credit_transaction.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    debit_from = models.ForeignKey(
+        'Organization',
+        related_name='debit_transactions',
+        on_delete=models.PROTECT,
+        null=True  # could be NULL if this is not a trade
+    )
+
+    value = models.DecimalField(max_digits=20,
```
klefler: Would it make sense to name this column **credit_value**? It would be more explicit... also, is "value" a reserved word in the database?

backend/api/models/credit_transaction_type.py
```diff
@@ -0,0 +1,18 @@
+from django.db import models
+
+class CreditTransactionType(models.Model):
```
klefler: Apologies in advance if I'm reading the class definition wrong, but this should have the Auditable, EffectiveDates, and Description columns.

klefler:
Hi @plasticviking (including @amichard here too)... I was looking at the relationship from the "credit_class" to the "sales_submission" table and it sparked a couple of thoughts. One of them was, "what's the difference between a credit class and a zev class? Are they the same? Should the table be renamed to zev_class_type"? (I added comments within the files during my review).

Then I started to wonder if we actually needed to store the zev class value in the vehicle table. Right now it isn't, but it looks like it would be calculated and displayed on the fly. Is there any value in storing this code in the vehicle table? I'm thinking there might be, especially when it comes to reporting/querying this table. It doesn't make sense to me to store this value in the sales_submission... because if we stored it with the vehicle we could then get it from the CSV template  (with the vehicle information) when the supplier submits their sales... that in itself would use existing data to derive the credit classes being requested... although it would mean extra joins to get the info. Anyways, I'm just trying to think where the best "root location" for the "class" would make sense, and it feels like the vehicle table is it, because it is already technically defined there (it's just not stored).

Sorry to bring up a potentially complex topic... please let me know if this has already been discussed and decided on.

backend/api/models/credit_class.py
plasticviking: As I understand it, these are different. ZEV credits can be (for now), A or B class. ZEV vehicles can be Battery-Electric, Fuel-Cell, etc. The formula for credit generation will depend on the type and range.

backend/api/models/credit_class.py
plasticviking: I omitted Auditable intentionally, since the intention of Auditable is to provide a record when rows are updated by application users. There is no provision in the application or design to have these tables be writable -- what's in them can only be modified by migration or ops_data scripts and hence has no meaninful user context.

backend/api/models/credit_transaction.py
plasticviking: I can do it that way, if you prefer. My personal preference is to avoid namespacing columns when the table is already suitably named.

`credit.value` is more intuitive to me than `credit.credit_value`.

`value` is not a reserved word (`values` is).

backend/api/models/credit_transaction_type.py
plasticviking: Again, the omission of `Auditable` was intentional since the contents of this table cannot be modified by application users.

`EffectiveDates` is perhaps negotiable. In my view, the purpose of effective dates is to filter selectable values presented to the user so deprecated or future values are not valid. It makes sense for things like a vehicle type, where we perhaps expire the ability to create certain classes of vehicle over time.

These credit transaction descriptions are not directly selectable, but controlled by application logic -- if they become irrelevant for some reason, corresponding code changes will be required to indicate that they should not be generated any longer. I can see the point of having the dates for reference in the database by users without the context of the application code, but work will have to be put into making sure these are synchronized.

We had similar columns in TFRS, and they went unused/empty.

plasticviking:
> Hi @plasticviking (including @amichard here too)... I was looking at the relationship from the "credit_class" to the "sales_submission" table and it sparked a couple of thoughts. One of them was, "what's the difference between a credit class and a zev class? Are they the same? Should the table be renamed to zev_class_type"? (I added comments within the files during my review).
> 
> Then I started to wonder if we actually needed to store the zev class value in the vehicle table. Right now it isn't, but it looks like it would be calculated and displayed on the fly. Is there any value in storing this code in the vehicle table? I'm thinking there might be, especially when it comes to reporting/querying this table. It doesn't make sense to me to store this value in the sales_submission... because if we stored it with the vehicle we could then get it from the CSV template (with the vehicle information) when the supplier submits their sales... that in itself would use existing data to derive the credit classes being requested... although it would mean extra joins to get the info. Anyways, I'm just trying to think where the best "root location" for the "class" would make sense, and it feels like the vehicle table is it, because it is already technically defined there (it's just not stored).
> 
> Sorry to bring up a potentially complex topic... please let me know if this has already been discussed and decided on.

Hi @klefler. I'm not sure I follow what you're getting at. `credit_class` and `vehicle_class_code` are both FKs in `vehicle`, referencing their respective lookup tables.

The `credit_class` and `credit_value` columns in `vehicle` can be derived based on business rules from the other values in `vehicle`, but are stored for convenience and for audit purposes.

backend/api/models/credit_transaction.py
klefler: Yes, thank you. The preference would be to use credit_value as a column name.

backend/api/models/credit_class.py
klefler: The intention of the auditable columns goes beyond users who update them, and should technically have no business meaning. If the system changes the values or if a script is run to update the values, we should be capturing the audit info. Audit columns should be on all tables. I am also suggesting we add them to keep a consistent pattern with our lookup tables, which is not just a ZEVA pattern, but one we use across NRM applications.

backend/api/models/credit_class.py
klefler: Yep, I think we might need the ZEV type (BEV, etc.) and then the associated ZEV Class (A, B)... I thought I saw a ZEV class (A or B) on the vehicle screen designs and heard this may be calculated on the fly (based on other attributes) and just displayed on the screen. If this is the case, I'd recommend that we store the calculated value (A, B) in the Vehicle table as well as the ZEV type (BEV, etc.) But if I misinterpreted this, then this table is fine as the "credit_class" table.... but I'd still like to see the table name change to "credit_class_code" or "credit_class_type" in that case.

backend/api/models/credit_transaction_type.py
klefler: I made some comments above re: audit columns and recommend adding them to this table.

Re: effective dates, I have two thoughts on why they should be added... one being as you mentioned above in that we should have the capability to "turn off" the code rather than deleting it... we want to be able to maintain referential integrity to the values that "once were" as converting data over to a new code is not typical (at least in my experience). The dates allow for that reference to be maintained for historical purposes, but allows the value to be ignored in future records.

The second thought I have is around keeping the lookup tables consistent, with the same columns and behaviours. I understand that there is extra logic in maintaining the dates and perhaps maybe a boolean column would make things easier? This is something that can be discussed, but it would mean the other lookup tables take a different approach to expiring their code values...

I am curious about why these columns are unused in TFRS... that wouldn't have been the intention in the design... but that's off topic and doesn't need explaining here.

klefler:
@plasticviking I think what I was referring to was "zev" class code (which we are calling credit_class at the moment).... I see there is a vehicle_class_code but (from what I understand) this is storing the type of vehicle class (two-door, mini compact, etc.) I made a comment under the credit_class file that may explain my thoughts on why we need the ZEV class code (A, B), which seems to be the same as the credit_class being proposed. Hope that helps clarify what I meant.

klefler:
@plasticviking - ok, I see the credit_class and credit_value columns in the table now... sorry if I missed them initially.

backend/api/models/credit_class.py
plasticviking: I've renamed this to `credit_class_code`

backend/api/models/credit_class.py
plasticviking: We do not update the audit tables when migrations or ops data scripts run. The only processes which can update them are user-interactive. If it is necessary to audit scritps, it should be added as a backlog request (and probably in TFRS too -- we don't do it there either). It would also be necessary to create a `system` user account or similar if you want the `user` audit fks to be something other than NULL.

backend/api/models/credit_transaction.py
plasticviking: Done

backend/api/models/credit_transaction_type.py
plasticviking: @klefler, I wasn't suggesting we do anything to break referential integrity (in fact, we have constraints that would block this), only that any kind of change that is going to render one of these values obsolete will necessarily have a corresponding code change. I get where you're coming from if these values were freely-selectable, but they are assigned based on a business rule, a key difference with some of the other lookup tables which do have potentially meaningful effective dates.

If a rule becomes obsolete and values of a certain class will no longer be referenced in new transactions, an application change is needed (probably with a hard-coded cutover date).

backend/api/models/credit_class.py
klefler: Ok thanks. Yes, something I will keep in mind for a future conversation.

PR #135, opened 2020-04-01
zeva 110 - upload icbc spreadsheet (front end)

emi-hi:


frontend/src/app/router.js
```diff
@@ -179,6 +179,11 @@ class Router extends Component {
                 path={ROUTES_CREDITS.LIST}
                 render={() => <CreditsContainer keycloak={keycloak} user={user} activeTab="transactions" />}
               />
+              <Route
+                exact
+                path={ROUTES_CREDITS.UPLOADVERIFICATION}
```
amichard, line 184: Non-blocking comment:
I'd suggest that this changes to UPLOAD_VERIFICATION for more readability.

frontend/src/credits/CreditsContainer.js
```diff
@@ -5,14 +5,28 @@ import React, { useEffect, useState } from 'react';
 import CreditTransactions from './components/CreditTransactions';
 import Loading from '../app/components/Loading';
 import CreditTransactionTabs from '../app/components/CreditTransactionTabs';
+import UploadVerificationData from './components/UploadVerificationData';
 import ROUTES_CREDITS from '../app/routes/Credits';
+import ROUTES_ICBCVERIFICATION from '../app/routes/ICBCVerification';
```
amichard, line 10: Non-blocking:
I would say it should be  ROUTES_ICBC_VERIFICATION so its more readable

PR #148, opened 2020-04-16
ZEVA 111 -  Back end component - parse uploaded file from ICBC

emi-hi:
-added model for icbc vehicle
-parses spreadsheet, drops unneeded columns
-checks make and year tables, adds ids to matching records from spreadsheet. Any missing get added to the tables
-checks icbc vehicle table for matching vehicles from spreadsheet (make, year, model), adds any missing
-adds all unique vins to registration table


backend/api/services/icbc_upload.py
```diff
@@ -0,0 +1,100 @@
...
+from api.models.vehicle_make import Make
+from api.models.icbc_vehicle import IcbcVehicle
+from api.models.model_year import ModelYear
+
+
+def insert_to_database(df, listName, table, columnName, requesting_user):
+    for each in listName:
+        try:
+            #get the id of that value from the table
+            theId = table.objects.get(name=each).id
```
amichard, line 13: theId is probably best written in snake case: the_id for consistency

backend/api/services/icbc_upload.py
```diff
@@ -0,0 +1,100 @@
+import pandas as pd
+from django.core.exceptions import ObjectDoesNotExist
+from api.models.icbc_registration_data import IcbcRegistrationData
+from api.models.vehicle_make import Make
+from api.models.icbc_vehicle import IcbcVehicle
+from api.models.model_year import ModelYear
+
+
+def insert_to_database(df, listName, table, columnName, requesting_user):
```
amichard, line 9: idealy, we make the variable a bit more descriptive, such as df into dataframe as its easier to follow

PR #153, opened 2020-04-23
ZEVA-134 view organizations list

emi-hi:
-added supplier details back to nav bar
- updated supplier list
-added tabs to individual suppliers (not working yet)

### DIFF
--- a/frontend/src/organizations/components/OrganizationsTable.js
+++ b/frontend/src/organizations/components/OrganizationsTable.js
@@ -18,11 +18,16 @@ const OrganizationsTable = (props) => {
     Header: 'Class',
     id: 'class',
   }, {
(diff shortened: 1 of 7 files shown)


frontend/src/organizations/components/OrganizationsTable.js
```diff
@@ -18,11 +18,16 @@ const OrganizationsTable = (props) => {
     Header: 'Class',
     id: 'class',
   }, {
-    className: 'col-compliance-requirement',
-    Header: 'Compliance Requirement',
+    accessor: () => '-',
+    className: 'col-credit-balance',
+    Header: 'A-Credits',
+    id: 'b-credits',
```
amichard, line 24: i think it might be better to change this into a-credits

PR #167, opened 2020-05-06
ZEVA-143: Create New Organization

emi-hi:
-BACKEND: added create serializer for new organizations
-FRONTEND: added checks in front end for whether organization is new (no id), and then renders form and axios post accordingly
- removed id from vehicle supplier tabs proptypes since new supplier will have no id, so that might need to be reviewed later

### DIFF
--- a/backend/api/serializers/organization.py
+++ b/backend/api/serializers/organization.py
@@ -4,6 +4,7 @@ from api.models.organization import Organization
 from api.serializers.organization_address import OrganizationAddressSerializer
 from api.models.organization_address import OrganizationAddress
 from datetime import date
+from symtable import Class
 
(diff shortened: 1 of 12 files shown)


backend/api/serializers/organization.py
```diff
@@ -4,6 +4,7 @@
 from api.serializers.organization_address import OrganizationAddressSerializer
 from api.models.organization_address import OrganizationAddress
 from datetime import date
+from symtable import Class
```
amichard, line 7: I'm not seeing that code that uses this. Do we still need this? What does Class do?

emi-hi: hmm sometimes when i'm typing it autocompletes and adds new imports.. I've caught it a couple times but must have missed one! I'll remove it now

PR #172, opened 2020-05-12
ZEVA-149: Imported roles and permissions from TFRS

amichard:
Changelog:
- Imported the Models
- Created a migration based on the roles and permissions that the keycloak is using


backend/api/models/permission.py
```diff
@@ -0,0 +1,39 @@
...
+from auditable.models import Auditable
+from api.managers.permission import PermissionManager
+
+
+class Permission(Auditable):
+    """
+    Contains the list of permissions to grant access to certain actions of
+    areas for the system.
+    """
+    code = models.CharField(
```
klefler, line 11: What are your thoughts on renaming this column "permission_code"? I see in TFRS it was called "code", but it would be a little more descriptive to change the name... but not if it's going to be a hassle.

backend/api/models/permission.py
```diff
@@ -0,0 +1,39 @@
...
+        max_length=100,
+        unique=True,
+        db_comment="Permission code. Natural key."
+    )
+    name = models.CharField(
```
klefler, line 16: Same note as "code"... would be a pain to name this column "permission_name"? I worry that "name" might be a reserved word too... so we'd want to add a more descriptive name to it, if it's not too much work.

backend/api/models/permission.py
```diff
@@ -0,0 +1,39 @@
...
+    objects = PermissionManager()
+
+    def natural_key(self):
+        """
+        Allows code to be used to identify a row in the table.
+        """
+        return (self.code,)
+
+    class Meta:
+        db_table = 'permission'
```
klefler, line 36: To keep this consistent with other "code" tables, we could call this "permission_code"... unless it's too much work.

backend/api/models/role.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def natural_key(self):
+        """
+        Allows us to match with the role name
+        """
+        return (self.name,)
+
+    # Add effective_date and expiration_date
+    class Meta:
+        db_table = 'role'
```
klefler, line 64: Similar thoughts as I had with the "permission" table... would it make sense to call this table "role_code" given that it's a list of codes?

And if we change it, would it make sense to rename the "name" column to either "role_name" or "role_code"?

backend/api/models/permission.py
amichard: I'll rename the column to permission_code

backend/api/models/permission.py
amichard: I'll rename this to permission_code

backend/api/models/permission.py
amichard: I think it makes more sense to keep the name permission as the idea behind the table is to contain the list of possible permissions and not just focus only on the code

backend/api/models/role.py
amichard: Same as above. Since the idea of the table is to contain the roles and i think it makes sense to keep the name

backend/api/models/permission.py
klefler: Ok, thanks Richard.

PR #253, opened 2020-07-16
ZEVA-199: Credit transaction model changes

amichard:
Changelog:
- Added weight class and model year to the credit transaction table
- Renamed fixtures so they're plural


backend/api/models/weight_class.py
```diff
@@ -0,0 +1,27 @@
...
+
+class WeightClass(EffectiveDates, Auditable):
+    """
+    A lookup table for weight classes. Initially, LDV only,
+    but with room to expand later.
+    """
+    class Meta:
+        db_table = "weight_class_code"
+
+    weight_class_code = models.CharField(
```
klefler, line 18: Just wondering if we need a Description column as well?

amichard: sure. done.

PR #282, opened 2020-08-18
ZEVA 212: Added checkbox to allow users to claim additional credit 

amichard:
Changelog:
- Added checkbox to claim additional credit for certain engine types
- Updated model so that it can save the claim for additional credit
- Checking additional credit shows upload files in the model creation form
- Added new calclulation


backend/api/models/vehicle.py
```diff
@@ -66,6 +66,11 @@ class Vehicle(Auditable):
         on_delete=models.PROTECT,
         null=True
     )
+    additional_credit = models.BooleanField(
```
klefler, line 69: Just wondering if the column name should be consistent with other boolean columns in the database. I think the convention used was 'is_' or 'has_' ... would 'has_additional_credit' make sense?

backend/api/models/vehicle.py
```diff
@@ -66,6 +66,11 @@ class Vehicle(Auditable):
...
+        default=False,
+        db_comment="Boolean Field to see whether the vehicle should get "
+                   "additional credit"
+    )
```
klefler, line 73: If this column is specific to only the US06 credit request, could this be added to the column comment?

i.e. "Boolean field used to claim whether the vehicle should get additional credit for passing the US06 range test."

backend/api/models/vehicle.py
amichard: initially tried that, but couldn't change "Can claim additional credit" into an "is/has" format

backend/api/models/vehicle.py
amichard: il change the comment

backend/api/models/vehicle.py
amichard: maybe "has_passed_us06_test" might make more sense

klefler: ya that or "has_claimed_us06_credit"?

PR #286, opened 2020-08-21
backend for transfer page now works

emi-hi:


emi-hi:
-added table for transfers with another status table
-can insert multiple rows or just one

PR #295, opened 2020-09-01
ZEVA 217 - supplier address form

emi-hi:
-added address type table to database
-added service and records addresses to supplier form
-suppliers can now have two addresses


backend/api/models/address_type.py
```diff
@@ -0,0 +1,27 @@
...
+"""
+from django.db import models
+
+from api.models.mixins.effective_dates import EffectiveDates
+from auditable.models import Auditable
+
+
+class AddressType(EffectiveDates, Auditable):
+    """
+    A lookup table for address type.
```
klefler, line 12: Table looks good.
Just one recommendation would be to add brief definition of the "records" and "service" address type so it's clear what they mean. Not sure if we can pull this out of the Act, i.e.

_"address for service is an address in British Columbia provided under section 17 (3) (d) [model year report] as an address at which notices and other communications will be accepted as served on or otherwise delivered to the supplier"_

klefler: Actually, the ticket has some easy definitions:

Service = the address for service of the supplier (although it's not clear what "service" means)
Records = the address where the supplier keeps records required to be kept and
maintained under the Act

PR #300, opened 2020-09-04
ZEVA 237 - front and backend tests

emi-hi:
created storybook tests for actions banners and vehicle save tests on backend


amichard (approved):
Looks good!

PR #302, opened 2020-09-09
transaction address bug

emi-hi:
added conditional to displaying records and service addres so that it doesn't give error


amichard (approved):
Looks good!

PR #317, opened 2020-09-16
icbc data current to: no data uploaded yet

emi-hi:
added no data uploaded yet to previousUpload beginning state


amichard (approved):
Looks good!

PR #322, opened 2020-09-17
bug 3 back button filtering

emi-hi:
added filtering when back button is clicked

"When you are in select ZEV model manufacturers, and then click on row to see details. You are brought to the details page which is nice. But then if you click 'baclk' to go back to your filtered search, you are instead brought back to the original list of all zev models. so you have to redefine your search criteria. Would be nice if when you select 'back' the zev model filter you applied sticks when you're switching from detailed view to filtered zev model view."


amichard (approved):
Looks good!

PR #323, opened 2020-09-17
bug 8 -  icbc upload upload date display

emi-hi:
added condition for icbc upload date display so it says either the date or "no icbc data uploaded yet"


amichard (approved):
Looks good!

PR #324, opened 2020-09-18
bug 21 - Validate / revalidate / recommend issuance buttons

emi-hi:
after a sale has been validated, the validate button will reappear on the page with a comment box and "recommend issuance" button--swapped order of these buttons and made the recommend issuance primary, changed button to "re-validate"


amichard (approved):
Looks good!

PR #326, opened 2020-09-21
bug 26 + testing

emi-hi:
-axios call and  banner links to transfer banners for idir and bceid users, banners filter transfer table now the same way that the other banners do
-friday testing


amichard (approved):
Looks good!

PR #328, opened 2020-09-21
Credit Transaction Request Refactoring

amichard:
Changelog:
- Merged Sales Submissions and Credit Transaction Requests
- Reduced the number of serializers used
- Removed unused pages and functions
- Improved displaying of error codes
- More consistent styling
- Code cleanup


emi-hi (approved):
I think there are going to be some conflicts with the dashboard work I was doing-- so if we merge it right away we should keep that in mind!

PR #331, opened 2020-09-23
ZEVA 278 - transaction list negative values

emi-hi:
added formatting for negative values (red)
fixed balance so it takes into account transfers (removes the amount from total)


amichard (approved):
Looks good!

PR #335, opened 2020-09-25
ZEVA 266 - vehicle model details page, form, and list table changes

emi-hi:
added alerts and comments
added stock comment for idir to request changes
reordered columns for vehicle model table, and removed some
disabled clear filter button until filter is selected


amichard (approved):
Looks good!

PR #337, opened 2020-09-29
zeva 271- consolidate button files

emi-hi:
removed individual button components
made one button component that holds submit, save, back, delete, and download
switched over instances where button html was written and instead imported button component
added check in vehicle min serializer for request.user


frontend/src/app/components/Button.js
```diff
@@ -0,0 +1,97 @@
...
+    case 'back':
+      onclick = () => { getRoute(); };
+      text = 'Back';
+      icon = 'arrow-left';
+      break;
+    case 'delete':
+      icon = 'trash';
+      text = 'Delete';
+      classname += ' text-danger';
+      onclick = () => { action(); };
```
amichard, line 31: You should be able to get away with just onclick = action;

This makes it simpler, so in case the action requires parameters (like passing e or element) then you don't have to manually pass it in here

amichard (approved):
Looks good!

PR #341, opened 2020-10-01
zeva 286 - credit application table

emi-hi:
reordered columns
changed some column headers
removed warnings for records that have sales (checked)
disabled button if no warnings (but still shows up for bceid user for issued)


amichard (approved):
Looks good!

PR #348, opened 2020-10-06
ZEVA 284 - status alerts

emi-hi:
added alert type for credits, fixed moment formatting for date time, added more storybook tests for alerts


amichard (approved):
Looks good!

PR #353, opened 2020-10-08
ZEVA 239 - credit application history

emi-hi:
added submission history model and serializer


backend/api/serializers/sales_submission.py
```diff
@@ -235,6 +236,11 @@ def update(self, instance, validated_data):
         validation_status = validated_data.get('validation_status')
 
         if validation_status:
+            SalesSubmissionHistory.objects.create(
+                submission=instance,
+                validation_status=validation_status,
+                update_user=request.user.username,
```
amichard, line 242: can we also include create_user here? just have them be the same as update_user

amichard (approved):
Looks good!

PR #356, opened 2020-10-13
zeva 285 - credit application summary table

emi-hi:
added summary table to credit application detail page
added org address for either idir or bceid view 
added comments into comment component
fixed styling on alert and comment


amichard (approved):
Looks good!

PR #366, opened 2020-10-22
ZEVA 192 - credit transfer other supplier accepts/rejects

emi-hi:
added transfer detail page
user can check off to confirm they are employee and accept transfer
user can comment on transfer and reject


amichard (approved):
Looks good!

PR #369, opened 2020-10-26
ZEVA 301- Change - only credit seller allowed to intitiaite transfer

emi-hi:
edited save function and form fields so that only credit seller can initiate transfer
-removed radio buttons
-hard coded debit from variable in save function in container


amichard (approved):
Looks good!

PR #371, opened 2020-10-27
ZEVA 307 - add optional company/attorney

emi-hi:
-added field on suppller address form for representative
-updated backend to store representative name in address table 

-currently not displaying name anywhere except on that form (it is not on the details page)


amichard (approved):
Looks good!

PR #377, opened 2020-10-29
ZEVA 236 - transfer form unfilled fields

emi-hi:
-disable checkboxes when row fields are not filled
-clear checkboxes when new field is added and then typed in
-tooltips on checkboxes and submit button when disabled
-blank rows removed before posting to database


amichard (approved):
Looks good

PR #378, opened 2020-10-29
ZEVA 282 - new activity banners for bceid

emi-hi:
-added rejected models and transfers activity banners for bceid


amichard (approved):
Looks good!

PR #382, opened 2020-10-30
Friday testing

emi-hi:
- fixed action bar in transfer details
- added tests
-alert now says "icbc date current to: no icbc data" if no icbc data uploaded, otherwise it will have the date


amichard (approved):
Looks good!

PR #384, opened 2020-11-05
ZEVA 318 - edit credit transfer draft

emi-hi:
-added edit ability by loading new form with prefilled values when draft record is clicked
-added serializer for updating transfer content and partner
-changed transfer container and page to be editContainer and form to match other components
-added check for partner and filled rows before saving


amichard (approved):
Looks good!

PR #390, opened 2020-11-10
zeva 300 - Analyst reviews transfer

emi-hi:
-removed 'draft' and 'submitted to partner' transfers from idir users list of transfers
-removed 'recommended' status transfers from bceid users list of transfers
-added transfer history table and serializer to backend to track status change
-added components for two details tables and action bar


amichard (approved):
Looks good

PR #392, opened 2020-11-12
Upload directory back from tmp

amichard:
Changelog:
- Temporarily removed automatic file deletion
- Changed folder back to the way it was


kuanfandevops (approved):
looks good

PR #395, opened 2020-11-13
Zeva 321

NavpreetGrewal:


frontend/src/credits/components/CreditTransfersDetailsSupplierTable.js
```diff
@@ -49,17 +49,27 @@ const CreditTransfersDetailsSupplierTable = (props) => {
     className: 'text-center',
     columns: [{
       headerClassName: 'd-none',
-      id: 'new-balance-a',
+      id: 'newA',
       accessor: (item) => (`${item.newBalanceA}-A`),
       className: 'text-right',
       width: 125,
+      getProps: (state, rowInfo) => ({
```
amichard, line 56: I'm assuming the idea behind this is to highlight the text as red.

Instead of using a style attribute my suggestion is to update the className instead
so I would so something like this:
className: `text-right ${(rowInfo.row.newB.slice(0,2) < 0) ? 'text-danger : ''}`

frontend/src/credits/CreditTransfersDetailsContainer.js
```diff
@@ -19,17 +19,39 @@ const CreditTransfersDetailsContainer = (props) => {
...
+    
     axios.get(ROUTES_CREDIT_TRANSFERS.DETAILS.replace(':id', id))
       .then((response) => {
         setSubmission(response.data);
+        orgId = response.data.debitFrom.id;
         setLoading(false);
       });
-  };
+    
+    axios.get(ROUTES_CREDIT_TRANSFERS.LIST).then((listresponse) => {
```
amichard, line 40: This might not fully get all the pending credit transactions so the count might be a bit off.

I noticed that this only retrieves credit transfers. we need to include credit applications as well

To clarify, an organization has insufficient credits if their current balance + credit applications (not counting drafts and issued) - credit transfers (not counting drafts and issued) - the current transfer application is less than 0

frontend/src/credits/CreditTransfersDetailsContainer.js
```diff
@@ -19,17 +19,39 @@ const CreditTransfersDetailsContainer = (props) => {
...
+      const organization_transfers = listresponse.data.filter((eachRecord) => eachRecord.debitFrom.id === orgId);
+      currentBalanceA = organization_transfers[0].debitFrom.balance.A;
+      currentBalanceB = organization_transfers[0].debitFrom.balance.B;
+      organization_transfers.forEach((each) => {
+        each.creditTransferContent.forEach((eachContent) => {
+          (eachContent.creditClass.creditClass === 'A') ?
+          transferContentA += eachContent.creditValue * eachContent.dollarValue : transferContentB += eachContent.creditValue * eachContent.dollarValue
```
amichard, line 47: I'd be more explicit and add another condition for creditClass B. As we need to, while very unlikely, ensure credit class C doesn't ruin the calculation

NavpreetGrewal:
@amichard Thanks for your feedback. I will work on it.

amichard (approved):
Looks good!

PR #398, opened 2020-11-13
ZEVA 324 - user role information toolip

emi-hi:
-added fixtures for role descriptions
-switched tables in frontend that display roles to use role code instead of description
-added tooltip to role info icon


amichard (approved):
Looks good!

PR #408, opened 2020-11-23
Zeva 344

NavpreetGrewal:
Add validation status "Rejected" to the condition to display comment for both bceId and government user.


amichard (approved):
Looks good!

PR #412, opened 2020-11-25
ZEVA 193 - Director approves transfer

emi-hi:
-non government users now see "approved by transfer partner" instead of "recommended approved" in transfer list view
-supplier balance table in transfer details is fixed
-when approve transfer is clicked:
    -backend checks initiating supplier totals (calculates balances by model year and type, and compares against each row in transfer) , if supplier does not have enough credits for any row the transfer will fail 
    - if transfer does not fail, a transaction record is added for each row in transfer
    - balances for each supplier are updated for A and B credits


amichard (approved):
Looks good!

PR #413, opened 2020-11-25
Zeva 325: BCEID Organization Administrator

NavpreetGrewal:


frontend/src/users/components/UserDetailsForm.js
```diff
@@ -23,6 +23,19 @@ const UserDetailsForm = (props) => {
     return <Loading />;
   }
 
+  const disableEditing = (role) => {
+    if (!role){
+      user.roles.forEach((eachRole) => { 
+        if (eachRole.roleCode === "Organization Administrator"){
```
amichard, line 29: I'd rather we check for permissions than actually checking for role code or name. These codes can potentially change.

NavpreetGrewal: oh ok I was not aware of that. I will make the required changes.

amichard (approved):
Looks good!

PR #414, opened 2020-11-25
ZEVA 342 - sales submission list table date field - use history table to grab submission date

emi-hi:
updated date field in sales submission to show submission date if the current status is government-altered ie recommended or checked and the user is BCEID


amichard (approved):
Looks good!

PR #427, opened 2020-12-04
ZEVA-356: Transfer error

emi-hi:
added more backend checks for transfers
-analyst will see a banner if this transfer will put supplier into negative (including pending)
-supplier cannot submit transfers that will put them into negative (including pending)
-director cannot issue credits if supplier does not have enough credits (backend only)


amichard (approved):
Looks good!

PR #437, opened 2020-12-09
Zeva 343:Display "New User" button and make user list clickable only if user is admin inside Vehicle supplier/users tab

NavpreetGrewal:


amichard:
Looks good. But can you please rename the fixture to use 0016 instead? Just to avoid conflicts in the number

amichard (approved):
Looks good!

PR #439, opened 2020-12-09
ZEVA-360: Credit Transfer Rescind Option

emi-hi:
-adds rescind status and buttons for either supplier to change status at any stage up to director issuing credits
-adds redirect for users attempting to edit transfers that are not in draft or rescind or they are not initiators
- adds information for signed by USER of COMPANY at DATE for bceid users (was previously just idir) and updated logic so if transfer has been rescinded it won't show sign dates for anything prior to that


amichard (approved):
Looks good!

PR #446, opened 2020-12-11
Zeva 364

NavpreetGrewal:
Updates: 
- Display "New Supplier" and "New User" button only for IDIR Admin.
- Display "Edit" button in supplier info page only for IDIR Admin and added permission check on routes as well.
- Display "Update ICBC data" tab only when user has permission to edit ICBC data.

Email Notification- IDIR:
- Added check for empty values in the email column for Email notification.
- Updated email template text, made it exact same as provided template.


amichard (approved):
Looks good!

PR #453, opened 2020-12-14
added "disapproved" to filter list for bceid transfer viewset

emi-hi:
added disapproved status to transfer list for bceid --so they can see anything that has been rejected by partner or them


amichard (approved):
Looks good!

PR #459, opened 2020-12-16
ZEVA 383  - Transfer history comment

emi-hi:
switched backend/database so comments are associated with history ids instead of transfer ids


amichard (approved):
Looks good!

PR #461, opened 2020-12-17
ZEVA 358 - Director reject comment

emi-hi:
added transfer comments for director to create
changed comment display on frontend so if there are comments they show up regardless of user, but backend only passes history and comments that the user is eligible to see


amichard (approved):
Looks good!

PR #465, opened 2020-12-18
ZEVA 383 - Transfer history comment

emi-hi:
changed comments to be associated with history instead of transfer id


amichard (approved):
Looks good!

PR #470, opened 2020-12-29
ZEVA 347 - Credit transfer status alerts

emi-hi:
-split alerts so that each feature has its own file to do the logic and call the basic alert component, updated all features with alerts
-made comment component use more shallow comment array
-added all transfer status alerts


amichard (approved):
Looks good!

PR #474, opened 2021-01-04
ZEVA 360 - Delete transfer

emi-hi:
-adds delete button to form actionbar (draft only)
-adds modal types, modal now shows text and button based on type
-excludes deleted status transfers from viewset


amichard (approved):
Looks good!

PR #479, opened 2021-01-05
ZEVA 391 -  idir admin only role

emi-hi:
changes link on dashboard in administration section
adds user prop to supplier tabs and checks permissions before showing tabs


amichard (approved):
Looks good!

PR #482, opened 2021-01-06
Zeva 376, 379 : Send email notification to Analyst and Director based on their subscription for credit transfer.

NavpreetGrewal:
- Updated Alerts for Notification List container
- Removed send email logic from credit transfer serializer and added it into the viewset method as emails were getting sent before the actual update


amichard (approved):
Looks good!

PR #485, opened 2021-01-06
ZEVA-381: Send notifications to the transfer partner based on their subscription for credit transfer

NavpreetGrewal:


amichard (approved):
Looks good!

PR #487, opened 2021-01-07
ZEVA 395 - BCEID admin only role

emi-hi:
removed navbar link and activity section from bceid users that dont have permissions for sales and transfers


amichard (approved):
Looks good!

PR #489, opened 2021-01-07
credit transaction tab on navbar

emi-hi:
added condition for displaying credit transaction tab in navbar to idir users


amichard (approved):
Looks good!

PR #494, opened 2021-01-08
Thursday Jan 8 Transfer bugs and changes

emi-hi:
1. When a BCeID user creates a transfer notice with multiple lines it is possible to skip over (not interact with) the $ value field and still submit. In this case the submit button is enables and the save button is disabled when it should be the other way round.
2. moved to Jira-404
3. IDIR user should not be able to see transfer notices that are rejected or rescinded before they are submitted to government
4. IDIR user briefly sees the BCeID transfer notice form page when they  click to view a rescinded transfer. Please change so they don't see that form.
5. When the IDIR Director is viewing a transfer notices the new balance of Tesla (the seller) appeared with many decimal places. Please change to limit the balance to 2 decimal places. 
6. The Director is receiving the 'Insufficient error' warning incorrectly (i.e when there are enough credits in the balance of the seller)
7. After a transfer has been recorded the BCeID users cannot click on the transfer to see the details - the line in the table is not clickable

### DIFF
--- a/backend/api/viewsets/credit_transfer.py
+++ b/backend/api/viewsets/credit_transfer.py
@@ -47,6 +47,7 @@ class CreditTransferViewset(
                 CreditTransferStatuses.SUBMITTED,
                 CreditTransferStatuses.DELETED,
                 CreditTransferStatuses.RESCIND_PRE_APPROVAL,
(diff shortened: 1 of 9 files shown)


amichard (approved):
Looks good!

PR #495, opened 2021-01-08
transfer details table credit quantity ceil to formatNumeric

emi-hi:
fixes rounding on transfer details table


amichard (approved):
Looks good!

PR #499, opened 2021-01-11
Notification fix

NavpreetGrewal:
-Moved send email logic for transfer to "send_email.py"
- added filters to send email only to interested organizations.
- Exclude the user who is taking action on the item.


amichard (approved):
Looks good!

PR #509, opened 2021-01-14
ZEVA-411: Fixed insufficient credits warning error for director

amichard:
Changelog:
- Changed the logic to check for insufficient credits


NavpreetGrewal (approved):

emi-hi (commented):
is sufficient credits right? Its using the supplier balance table for calculating.. when we actually validate the transfer it looks at the supplier totals which uses the credit_transaction table

amichard:
You're right! It shouldn't count the transfers. Let me change the code

emi-hi (approved):

PR #519, opened 2021-01-20
Zeva 385, 386, 366 - Evidence

emi-hi:
frontend: allows user to select file from computer to upload, allows user to view and download said file
backend: serializers and viewset to save files to minio and info to database
database: model for sales evidence


amichard (approved):
Looks good!

PR #537, opened 2021-01-29
ZEVA 426 - evidence reupload options, deleting, and text on file types/numbers

emi-hi:
-adds checkbox to excel file upload to display or not (optional and will not throw errors if no file is selected
-adds excel file name so user knows what is already there
-adds ability to delete evidence or add new
-adds text for file types and max number for evidence


amichard (approved):
Looks good!

PR #540, opened 2021-01-29
ZEVA-241: Compliance Ratios Page

NavpreetGrewal:
Changelog:

- Added DB table compliance_ratio with columns model_year, compliance_ratio and zeva_class_a.
- Added fixture and serializer 
- Added endpoint to get the ratio list from the table
- UI changes to show the ratio list in the react table.
- Sorting is disabled on all the columns but this can be enabled if required.


frontend/src/compliance/ComplianceRatiosContainer.js
```diff
@@ -3,14 +3,32 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import CustomPropTypes from '../app/utilities/props';
 import ComplianceTabs from '../app/components/ComplianceTabs';
+import COMPLIANCE from '../app/routes/Compliance';
```
emi-hi, line 6: should this be ROUTES_COMPLIANCE?

emi-hi (approved):

frontend/src/compliance/ComplianceRatiosContainer.js
NavpreetGrewal: yes, you are right. I will fix it. Thanks!

PR #541, opened 2021-02-01
Changed field type to float

NavpreetGrewal:


emi-hi (approved):

amichard (approved):
Looks good!

klefler:
May I please ask a couple of questions regarding the new table? Just wondering if "model_year" is using a foreign key from the model_year table, or is this different?

Are there comments for the new _compliance_ratio_ and _zev_class_a_ columns? From a metadata perspective, primarily on the zev_class_a column, it should be clear what the purpose of the column is.

For example; 
zev_class_a = "_Contains the compliance ratio percentage for ZEV Class A vehicles in the light-duty motor vehicle class._"

Thanks!

NavpreetGrewal:
> May I please ask a couple of questions regarding the new table? Just wondering if "model_year" is using a foreign key from the model_year table, or is this different?
> 
> Are there comments for the new _compliance_ratio_ and _zev_class_a_ columns? From a metadata perspective, primarily on the zev_class_a column, it should be clear what the purpose of the column is.
> 
> For example;
> zev_class_a = "_Contains the compliance ratio percentage for ZEV Class A vehicles in the light-duty motor vehicle class._"
> 
> Thanks!

Sure, model_year is not using FK from model_year table at this point because in model_year table we have only 3 year (2019,2020, 2021) but for this new table, we need to show compliance ratios for all the years from 2019 to 2040. so thats why for now, I am not making this FK column. 

And yes you are right about comments, I think I added them initially but they got removed with my new PR. I will add them back. 

please feel free to reach out to me on discord if you have any further questions or concerns. Thanks!

klefler:
Thanks for the response @NavpreetGrewal. Sounds good! I saw the recent updates to the 2 columns and they look great.

PR #548, opened 2021-02-04
ZEVA 37 - compliance calculator page

emi-hi:
-frontend only
-calculator on top for seeing ratios per calendar year based on supplier size
-table on bottom renders list of vehicle models with credit amounts per vehicle, user enters number into input and it displays total credits per model based on quantity sold
-sum at bottom of page shows total a and b credits based on input


amichard (approved):
Looks good!

PR #558, opened 2021-02-09
ZEVA 444 - Ids CA and CT to alerts and tables

emi-hi:
adds id to alerts and list tables for credit transfers and applications


amichard (approved):
Looks good!

PR #559, opened 2021-02-10
ZEVA-457: Model Year Report - Index Page (front end)

NavpreetGrewal:
* Added new text
* Changed tab and heading text to Model Year Reports 
* Changed maxWidth so that it looks good on desktop screen.


amichard (approved):
Looks good!

PR #570, opened 2021-02-10
ZEVA-454:Create database tables to support backend for consumer sales

NavpreetGrewal:
Add ldv_sales column to model_year_report container model
Add model_year_report_vehicle model
Add model_year_report_previous_sales model


amichard (approved):
Looks good!

PR #572, opened 2021-02-11
ZEVA-464:Credit transfer pages - staying on transfer detail page after save/submit/rescind/recommend//reject/record

NavpreetGrewal:
If status is DRAFT, RESCIND, redirect user to edit page else reload the detail page to update the status alert.

### DIFF
--- a/frontend/src/credits/CreditTransfersEditContainer.js
+++ b/frontend/src/credits/CreditTransfersEditContainer.js
@@ -125,7 +125,14 @@ const CreditTransfersEditContainer = (props) => {
         debitFrom,
         signingConfirmation: checkboxes,
       })
-        .then(() => history.push(ROUTES_CREDIT_TRANSFERS.LIST))
+        .then((response) => {
+          if (status = "DRAFT") {
+            history.push(ROUTES_CREDIT_TRANSFERS.EDIT.replace(/:id/gi, response.data.id))
+          }
+          else {
+            history.push(ROUTES_CREDIT_TRANSFERS.DETAILS.replace(/:id/gi, response.data.id))
+          }
+        })
         .catch((error) => {
           const { response } = error;
           if (response.status === 400) {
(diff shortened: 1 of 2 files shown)


frontend/src/credits/CreditTransfersEditContainer.js
```diff
@@ -140,7 +147,13 @@ const CreditTransfersEditContainer = (props) => {
         debitFrom,
         signingConfirmation: checkboxes,
       })
-        .then(() => history.push(ROUTES_CREDIT_TRANSFERS.LIST))
+        .then(() => {
+          if (status == "DRAFT") {
```
amichard, line 151: Instead of doing a force reload to the whole page, you can try doing this instead:
```
          history.push(ROUTES_CREDIT_TRANSFERS.DETAILS.replace(/:id/g, id));

          if (status === 'DRAFT') {
            history.push(ROUTES_CREDIT_TRANSFERS.EDIT.replace(/:id/g, id));
          }
```
It kind of looks awkward, but it is one of the few ways to do a "force reload" without reloading the whole page.
Doing a window.location.reload(), causes the whole page to completely reload. It's not too bad, but it's inconsistent with the rest of the site

frontend/src/credits/CreditTransfersEditContainer.js
```diff
@@ -125,7 +125,14 @@ const CreditTransfersEditContainer = (props) => {
         debitFrom,
         signingConfirmation: checkboxes,
       })
-        .then(() => history.push(ROUTES_CREDIT_TRANSFERS.LIST))
+        .then((response) => {
+          if (status = "DRAFT") {
```
amichard, line 129: I think you might want to change this to "===" instead of a single "=". Single equal might not be checking for your condition

amichard (approved):
Looks good!

PR #578, opened 2021-02-19
ZEVA 447 - make models Inactive

emi-hi:
-adds a button on zev details page to make model inactive, which will change the alert that is displayed 
-adds field on zev list table for active status and ability to filter by this
-filters compliance calculator models by validation status and active status
-removes non active models from sales template


amichard (approved):
Looks good!

PR #590, opened 2021-02-25
ZEVA-476:Model Year Compliance Report - Confirmations

NavpreetGrewal:
- added fixture to update assertion table with compliance report checkboxes.
- Updated SigningAuthorityAssertionViewSet, removed module="credit_transfer" filter and added filtering on the front end logic.
- Getting all checkboxes from the DB table and displaying them on different pages based on module value.
-  Saving checkbox confirmation values for supplier_information and consumer_sales pages.
(need to move checkbox logic from credit activity to new compliance obligation)


amichard (approved):
Looks good!

PR #595, opened 2021-03-02
ZEVA-461,477:Model Year Report - Supplier Info and Consumer Sales page changes

NavpreetGrewal:
-Added input boxes for previous year sale instead of labels.
-saving input values to the DB.
-calculating average of 3 previous year values and supplier class based on the average.


amichard (approved):
Looks good!

PR #601, opened 2021-03-06
ZEVA-484: Change comment feature on credit application

NavpreetGrewal:
- Added react-quill package for comment editor. based on users formatting, it generates the html. I am saving html to the DB as a string.
- Retrieving comments from the DB and using html-react-parser to display the comment properly on screen.
- Added separate button to add comments and removed comment logic linked with other buttons.
(Main logic is in line 58-84, 213-248 and few other lines, Sorry for all these formatting changes, my formatter is not working properly)


amichard (approved):
Looks good!

PR #625, opened 2021-03-23
Changed permission for Zev Model notifications. Only zev_user will be able to see zev Model notifications.

NavpreetGrewal:
- Added new fixture to update the permission for ZEV Model notifications.


amichard (approved):
Looks good!

PR #627, opened 2021-03-24
Zeva 500 - Compliance Obligation Ratio reduction

emi-hi:
The Ratio Reduction table on the Compliance Obligation page contains:
The total LDV Sales that was input and saved on the Consumer Sales tab
The Compliance Ratio is pulled from DB for the MY of the report
If the supplier is classed as Large then the Large Volume Supplier Class A Ratio is pulled from DB also, otherwise this line is not displayed
The Ratio Reduction is then calculated using these values
If the supplier is classed as Large then the 2 sub-bullets are displayed with calculated values
The ZEV Class A Debit is calculated using the Large Volume Supplier Class A Ratio
The Unspecified ZEV Class Debit is the remainder

-lints
-adds year field to model year report serializer
-adds proptypes
-adds format numeric to credits issued table


NavpreetGrewal (approved):

amichard (approved):
Looks good!

PR #632, opened 2021-03-25
Changed code to update_or_create so it updates the models when there are changes from the csv

amichard:
Changelog:
- Update the model and upload date as needed


kuanfandevops (approved):
Looks good, the changes are same a PR #630

PR #636, opened 2021-03-30
 ZEVA-498:Consumer Sales page - Retrieve values from the saved snapshot

NavpreetGrewal:
* Based on the report id, load the snapshots of previous_sales, vehicle_list, ldv_sales and confirmation if it exist in the DB, 
   else load vehicle_list from vehicle_sales, and display input boxes for previous sales & ldv_sales. 
* Made some minor name changes and formatting

### DIFF
--- a/backend/api/viewsets/model_year_report_consumer_sales.py
+++ b/backend/api/viewsets/model_year_report_consumer_sales.py
@@ -160,26 +160,3 @@ class ModelYearReportConsumerSalesViewSet(mixins.ListModelMixin,
             'organization_name': request.user.organization.name,
             'validation_status': report.validation_status.value,
             })
-
-    @action(detail=True)
-    def previouse_sales(self, request, **kwargs):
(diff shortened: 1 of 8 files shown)


backend/api/viewsets/model_year_report_consumer_sales.py
```diff
@@ -35,10 +53,11 @@ def create(self, request, *args, **kwargs):
         previous_sales = request.data.get('previous_sales')
         confirmations = request.data.get('confirmation')
         supplier_class = request.data.get('supplier_class')
+        previouse_years_exist = request.data.get('previous_years_exist')
```
amichard, line 56: not exactly incorrect code, but looks like a typo for the word previous

NavpreetGrewal: oh good catch, let me fix it.

amichard (approved):
Looks good!

PR #638, opened 2021-03-31
ZEVA 507 - compliance obligation save to database

emi-hi:
Addition of the back-end support required to all ow the user to save the information they review or enter in the compliance obligation page.


amichard (approved):
Looks good!

PR #655, opened 2021-04-15
ZEVA 502- retrieve values for summary page

emi-hi:
-changes obligation serializer to use id instead of year (also updates frontend)
-adds axios calls to summary container for report, supplier, consumer sales, and obligation detsails
-reworks how summary tables receive data (format has changed since frontend was originally developed)


amichard (approved):
Looks good!

PR #657, opened 2021-04-16
Compliance obligation cleanup

emi-hi:
-deletes old obligation snapshots before creating new ones
-loads and saves confirmations for obligation


amichard (approved):
Looks good!

PR #669, opened 2021-04-23
Applying PR668 to current DEV

amichard:
Changelog:
- Applying the same fix as #668


kuanfandevops (approved):
looks good, same changes as PR 668

PR #684, opened 2021-05-05
zeva 565 Summary offset and provisional numbers

emi-hi:
-adds provisional BEFORE offset and provisional AFTER offset variables (before includes pending, after is that plus offset)
-fixes an issue with years for ratios comparing string to int


amichard (approved):
Looks good!

PR #685, opened 2021-05-05
ZEVA-571:Fix read-only logic when a page is confirmed, also fixed some issues with confirmation retrieval on summary page.

NavpreetGrewal:
- Make everything read-only when status is confirmed or submitted.
- created new method for retrieving the summary confirmation.
- fixed a bug with saving summary confirmation.


amichard (approved):
Looks good!

PR #690, opened 2021-05-11
ZEVA-554:Update the Logic for Prior Credit Balance in Consumer Sales

NavpreetGrewal:
- Updated models 
   - changed ldv_sales field type from decimal to integer
- added logic to viewset for previous sales retrieval and finding a average sales based on the numbers.


amichard (approved):
Looks good!

PR #693, opened 2021-05-14
ZEVA-563: Add function for retrieving obligation balances broken down  by year

NavpreetGrewal:
- Created a function to calculate prior year balance

** pending: current year balances calculation needs more discussion.


emi-hi (commented):
prior year balance should be using the previous years ending balance

emi-hi (approved):

PR #712, opened 2021-05-28
Apply docker secret and update Jenkins base image

kuanfandevops:


kuanfandevops:
close without merging

PR #719, opened 2021-06-01
Model year report supplier information 3 previous years fix

emi-hi:
-filters by 3 prior years on backend for confirmed MY reports


amichard (approved):
Looks good!

PR #747, opened 2021-06-17
ZEVA 617 - assessment bugs - radio descriptions and report year

emi-hi:
-sets modelyear in assessment container
-adds check for decision descrition
-moves deficit code up and passes it with assessment even if no assessment exists (was previously passing nulls for everything)


amichard (approved):
Looks good!

PR #750, opened 2021-06-22
ZEVA-610:Summary Bugs

NavpreetGrewal:
- created a new param for summary page requests, if param == True, load from the snapshot.
- made bullet points consistent
-  Added message about last time saved and if page is not saved.


amichard (approved):
Looks good!

PR #751, opened 2021-06-22
Add volume for database

kuanfandevops:


kuanfandevops:
I'm not very sure if the database volume is needed, it may reduce migrations every time run compose up.

amichard (approved):

amichard:
We can give it a try. I tend to avoid them since it requires you have to postgres installed in your actual system. It shouldn't have to remigrate every time, unless you have to do docker-compose down.

PR #763, opened 2021-06-30
ZEVA 621 - If report is assessed, make the report read-only

emi-hi:
-conditionally removes any inputs if report has been assessed


amichard (approved):
Looks good!

PR #764, opened 2021-06-30
zeva 620-  credit reduction red font

emi-hi:
-adds red font color and minus to reduction values if they are greater than 0


amichard (approved):
Looks good!

PR #766, opened 2021-07-05
ZEVA-622: Edits from Analyst to LDV Sales need to be saved

NavpreetGrewal:
- added from_gov flag field to the obligation table
- new update method to add obligation changes from the govt to the DB


amichard (approved):
Looks good!

PR #769, opened 2021-07-07
ZEVA 598 - credit agreements frontend form

emi-hi:
-frontend for credit agreements form

-still needs to retrieve values for transaction types from backend
-can add axios post for comments and saving form after backend is finished


amichard (approved):
Looks good!

PR #781, opened 2021-07-12
ZEVA 600 - Credit agreement feature - document upload

emi-hi:
- adds update function in save serializer for adding attachment
-starts adding some other base serializers such as get update user
-removes history.push after agreement attachment upload and instead console logs (details page not built)

-document is uploaded when agreement is submitted NOT when upload button is clicked


amichard (approved):
Looks good!

NavpreetGrewal (approved):

PR #783, opened 2021-07-13
Zeva-595 add comment

kuanfandevops:
The comment can be added to the list. After the comment is added, the text in <ReactQuill> need to be cleared somehow. Also the id need to be unique in the comment list. This PR is for discussion only for now.

### DIFF
--- a/frontend/src/creditagreements/CreditAgreementsEditContainer.js
+++ b/frontend/src/creditagreements/CreditAgreementsEditContainer.js
@@ -39,19 +39,22 @@ const CreditAgreementsEditContainer = (props) => {
   const [errorMessage, setErrorMessage] = useState(null);
   const analystAction = user.isGovernment
   && user.hasPermission('RECOMMEND_COMPLIANCE_REPORT');
-  const handleCommentChangeIdir = (text) => {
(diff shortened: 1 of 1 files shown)


kuanfandevops:
Hi @emi-hi , need you advice that how to clear the comment after it is added to the list and how to create the unique id for the  comment list.

emi-hi:
> Hi @emi-hi , need you advice that how to clear the comment after it is added to the list and how to create the unique id for the comment list.

So since the layout is changing a bit and we are not going to use the comment array anymore i'll just give a brief explanation of what I would do
for ids, these are generated automatically (with timestamps) so no need to do that. I guess for this case since we aren't actually creating the comment right away i should have instructed you to not have anything in the id or timestamps.
For clearing the comment:
 i would set the default value for the input box as to a state like idirCommentInput, then when the 'add' button is clicked, i would add that  value to an idir comment array state and then set the idirCommentInput state to a blank value. 

Richard or I can give you an overview of state at some point if you'd like!

kuanfandevops:
I'll just close the PR for now as there may be some changes for the page. Also I'll keep the branch for a while in case I'll come back to it.

PR #785, opened 2021-07-15
zeva 629 - agreement comments backend save

emi-hi:
-adds field for optional agreement id to agreement model
-adds comment to agreeent save serializer
-adds comment to submit function data on frontend
-adds serializer method field for comments in agreementserializer


amichard (approved):
Looks good!

PR #790, opened 2021-07-19
Zeva 635: Initial PR for IDIR Display page

NavpreetGrewal:
- Added conditions in display page so that it display information on the page based on the user(supplier or Govt)
- if there is agreement id, get the details from the database which is working for idir user display page.
- if there is no id, such as for bceid user at this point, displaying static value( this will change as there will be ids list for bceId users once the transaction are complete, there is a new card 639 for this  in the backlog )
- made some changes in the serializer to get the details from the DB.( pending: attachment and credits data)
- Made UI changes.


amichard (approved):
Looks good!

PR #808, opened 2021-07-23
BugFix - 7,8,9,10

NavpreetGrewal:
**7.** Added blank line after Credit A Reduction heading if there is no reduction data present
**8.** Inconsistency in the behavior for what happens when the user changes LDV sales number. Sometimes the A credit reduction changes without saving and other times the user has to save first - **removed the condition to calculate reduction only if credit reduction selection is there inside onchangeLdvSales function. Now user can see reduction numbers changing without saving ldvSales first**
**9.** The text for all types of balances is all caps
**10.** The unspecified credit reduction appears to default to B without the user selecting either radio button. As a result the user can currently progress to finalize the report without selecting A or B.- **added condition, radioId should be present before calculation of unspecified reduction.**

### DIFF
--- a/frontend/src/app/utilities/calculateCreditReduction.js
+++ b/frontend/src/app/utilities/calculateCreditReduction.js
@@ -102,67 +102,69 @@ const calculateCreditReduction = (
   });
 
   remainingReduction = 0;
+ 
+  if (radioId) {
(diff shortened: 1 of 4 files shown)


amichard (approved):
Looks good!

emi-hi (approved):

PR #813, opened 2021-07-26
credit agreements linting

emi-hi:
-skipped list files due to potential conflicts but the rest is linted
-added some proptypes but more needs to be done


amichard (approved):
Looks good!

PR #826, opened 2021-08-05
ZEVA-636:Logic for agreement/adjustment transaction date

NavpreetGrewal:
- changed auto_now_add to False  for transaction_timestamp field and set default todays date if no specific date is provided
- update adjusts_credit for credit agreement


emi-hi (approved):

backend/api/models/credit_transaction.py
```diff
@@ -52,7 +53,8 @@ class CreditTransaction(Auditable):
     transaction_timestamp = models.DateTimeField(
         blank=False,
         null=False,
-        auto_now_add=True,
+        default=datetime.now(),
```
amichard, line 56: I'll have to test this, but using datetime.now() (with the parens means that I think it will be evaluated when you created the migration. So that's why it's hard-coding 2021/8/5). I think you should pass it without the parens (just datetime.now)

NavpreetGrewal: Thanks, its fixed now.

PR #828, opened 2021-08-05
Zeva633 & Zeva626 - optimize Dashboard and add alerts for credit adjustments

emi-hi:
-puts all the dashboard values into a serializer so the container isn't doing multiple database calls
-switches around frontend so that it uses the return from that single call to fill all the necessary fields
-adds adjustments to idir and bceid action files


amichard (approved):
Looks good!

PR #829, opened 2021-08-06
bug fix user edit and linting

emi-hi:
-fixes error with role.code givign error on user serializer
-lints


amichard (approved):
Looks good!

PR #842, opened 2021-08-12
Zeva656/651 - supplier information

emi-hi:


frontend/src/supplementary/SupplementaryContainer.js
```diff
@@ -9,33 +9,61 @@ const SupplementaryContainer = (props) => {
...
+    setNewData({ ...newData, comment });
   };
 
   const handleSubmit = (status) => {
-    const data = {
-      ...details,
-      status,
-    };
-    axios.patch(ROUTES_SUPPLEMENTARY.SAVE.replace(':id', id), data);
+    setNewData({ ...newData, status });
```
amichard, line 20: setNewData is fine here, but won't update newData right away, so I wouldn't rely on it. The console.log can be misleading as well, I'd check the headers and see what's actually getting passed.

I'd do something like this:
```
const data = {
  ...newData,
  [name]: value
};

setNewData(data);

axios.patch(ROUTES_SUPPLEMENTARY.DETAILS.replace(':id', id), data);
```

amichard (approved):
Looks good!

NavpreetGrewal (approved):

PR #844, opened 2021-08-13
ZEVA-661 and BUGFix: 

NavpreetGrewal:
-created supplemental report credit activity model
-BUGFIX-ZEVA user not able to see and click confirmation checkboxes on supplier info, consumer sales and credit activity page.
-BUGFIX-BCeID user can currently submit the report again after it has been recommended to the Director. This should be disabled/not shown the same as when the report is in submitted status
-BUGFIX-IDIR side: the Director can no longer edit the message to the supplier. This was working previously.


emi-hi (approved):

frontend/src/compliance/components/AssessmentDetailsPage.js
```diff
@@ -366,19 +366,20 @@ const AssessmentDetailsPage = (props) => {
...
-                    handleCommentChange={handleCommentChangeBceid}
-                    title="Assessment Message to the Supplier: "
-                  />
                 </div>
               </div>
             </div>
           </div>
         </>
       )}
+      {user.isGovernment && ( <CommentInput
```
amichard, line 375: if the only condition is just user.isGovernment, this will make it always show up if the user is an IDIR user, regardless of status. So even if the report has been assessed this will show up

PR #890, opened 2021-09-22
zeva 677 - bceid history

emi-hi:
-removes status change for bceid recommended, because it causes it to show up twice on list
-adds styling for history list
-puts history into seperate component


amichard (approved):
Looks good!

PR #902, opened 2021-09-29
ZEVA 698 - negative balances brackets

emi-hi:
-adds extra argument and logic to formatnumeric to show brackets instead of minus symbol for negative balances

-adds extra argument in formatNumeric call to spots where balance is listed

### DIFF
--- a/frontend/src/app/utilities/formatNumeric.js
+++ b/frontend/src/app/utilities/formatNumeric.js
@@ -13,7 +13,7 @@ const formatNumeric = (value, decimals = 2, roundBracket) => {
   }
   if (roundBracket) {
     if (newValue < 0) {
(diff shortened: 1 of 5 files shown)


NavpreetGrewal (approved):

frontend/src/app/utilities/formatNumeric.js
```diff
@@ -11,13 +11,18 @@ const formatNumeric = (value, decimals = 2) => {
   } else if (decimals === 0) {
     newValue = Math.round(newValue);
   }
-
+  if (roundBracket) {
+    if (newValue < 0) {
+      newValue = `(${Math.abs(newValue)})`;
```
amichard, line 16: Just double-checking. Math.abs will remove the decimals. Is this the intention?

PR #935, opened 2021-11-01
zeva 721 - delete supplementary draft bug

emi-hi:
-adds redirect to listing page if supplemental draft is deleted


amichard (approved):
Looks good!

PR #939, opened 2021-11-02
zeva723 - saved comments not appearing

emi-hi:
-adds exclude to supplemental  queryset to exclude deleted statuses
-adds bceid comment array to details page
-switches what bceid comment input is expecting for default value so comments actually appear


amichard (approved):
Looks good!

PR #1042, opened 2022-09-08
zeva-429--credit-agreements-non-integer-values

tim738745:


AlexZorkin (approved):
Looks good, let's just make sure existing data in test get converted over to floats correctly.

PR #1193, opened 2022-10-28
 feat/fix - zeva 467 - Create independent reasseement process for government

emi-hi:
feat: disables 'create new supplemental report' button if reassessment exists
fix: enables government user to create reassessment even if supplemental does not exist


emi-hi:
new wireframe was added today so i need to make some changes before it gets merged

frontend/src/compliance/AssessmentContainer.js
```diff
@@ -157,6 +158,11 @@ const AssessmentContainer = (props) => {
                 supplementaryResponse.data &&
                 supplementaryResponse.data.status
               ) {
+                setReassessmentExists(
+                  supplementaryResponse.data.reassessment.reassessmentReportId
```
AlexZorkin, line 162: Does data.reassessment always exist? If it could be null we need to add a question mark after reassessment to prevent a null error.

frontend/src/compliance/components/AssessmentDetailsPage.js
```diff
@@ -325,20 +326,21 @@ const AssessmentDetailsPage = (props) => {
                   (supplementaryStatus === 'DRAFT' && createdByGov) ||
                   supplementaryStatus === 'DELETED' ||
                   supplementaryStatus === 'ASSESSED') && (
-                  <button
-                    className="btn button primary"
-                    onClick={() => {
+                  <Button
+                    buttonTooltip="at this time its not possible as government has begun a reassessment"
```
AlexZorkin, line 330: Does this tooltip show even when reassessmentExists if false?

AlexZorkin (commented):
Are there any issues with creating a reassessment without having created a supplemental report before hand?

frontend/src/compliance/components/AssessmentDetailsPage.js
emi-hi: nope!

emi-hi: Ive moved it into the container and it now gets set only when reassessment exists, it wouldnt show up anyways but now its a blank string unless reassessment exists and then it gets that text

frontend/src/compliance/AssessmentContainer.js
emi-hi: I have added this to a condition

AlexZorkin (approved):

PR #1234, opened 2022-12-13
fix sorting

vibhiquartech:


AlexZorkin (changes requested):
Please take a look at the Administrative Credit Reduction transaction type in the transactions list. It is not showing the correct amounts in the balance column. Also please check the purchase agreement transaction type for correct balances using test data. Prod does not have any purchase agreement transactions yet but may in the future.

Last thing, when submitting prs please follow the PR naming conventions ex. Fix: Transaction Table Sorting - ZEVA 492

vibhiquartech:
Please verify the updated changes @AlexZorkin

frontend/src/credits/components/CreditTransactionListTable.js
```diff
@@ -186,7 +186,31 @@ const CreditTransactionListTable = (props) => {
       });
     }
   });
+  transactions.sort((a, b) => {
```
AlexZorkin, line 189: These two sorts can be combined and then remove the console.log(transactions) and we're good to go. thanks

AlexZorkin (changes requested):

AlexZorkin (approved):

PR #1251, opened 2022-12-15
fix - ZEVA 509 - Supplementary report - can't null out fields

emi-hi:
-catches null values for sales, credit activity, etc on the frontend and replaces them with the original values from the model year report when posting to supplemental_report_credit_activity

### DIFF
--- a/frontend/src/supplementary/SupplementaryContainer.js
+++ b/frontend/src/supplementary/SupplementaryContainer.js
@@ -273,7 +273,6 @@ const SupplementaryContainer = (props) => {
 
   const handleSupplementalChange = (obj) => {
     let creditActivity = [];
-
     if (newData.creditActivity) {
       ({ creditActivity } = newData);
     } else {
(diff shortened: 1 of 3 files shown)


frontend/src/supplementary/SupplementaryContainer.js
```diff
@@ -436,13 +471,17 @@ const SupplementaryContainer = (props) => {
...
+            dataToAdd.oldData = each.oldData 
+          }
+          zevSales.push(dataToAdd);
         }
       });
       newData.zevSales = zevSales;
-      if (forceSetNewData) {
-        setNewData({ ...newData });
-      }
+      // if (forceSetNewData) {
```
AlexZorkin, line 482: do we need to keep this here for reference?

AlexZorkin (changes requested):
Looks good, nice frontend solution to resetting those fields. Just that one cleanup line.

AlexZorkin (approved):

PR #1252, opened 2022-12-20
ZEVA 570 - Access evidence document issue fixed

vibhiquartech:


AlexZorkin (changes requested):
This wouldn't fix anything on test or prod so please take another look at this ticket. Please investigate the document links and see if the files are missing or mis-named etc.

PR #1329, opened 2023-01-27
ZEVA - 1288 BCeID user Credit Applications dashboard count fix

vibhiquartech:


emi-hi (dismissed):
the card is to fix the number for credit applications, not credit agreements. Also, removing the check for government user will not fix the issue because we do not want bceid users to see everything idir users see. I would suggest looking at what is passed from the backend to see if the issue is a frontend or backend issue.

AlexZorkin (changes requested):
Your idea will work but not in this form. You need to add another validation_status to the CreditRequestCountSerializer. Make sure you include 'CHECKED' as one of the validation statuses. 

class CreditRequestCountSerializer(ModelSerializer):
    status = SerializerMethodField()
    total = SerializerMethodField()

    def get_total(self, obj):
        return obj['total']

    def get_status(self, obj):
        request = self.context.get('request')
        print(obj['validation_status'].value)

        if not request.user.is_government:
            if obj['validation_status'].value == 'RECOMMEND_APPROVAL' or \
              obj['validation_status'].value == 'RECOMMEND_REJECTION' or \
                obj['validation_status'].value == 'CHECKED':
                    return 'SUBMITTED'
        return obj['validation_status'].value

    class Meta:
        model = SalesSubmission
        fields = ('status', 'total')

backend/api/serializers/dashboard.py
```diff
@@ -258,12 +259,17 @@ def get_activity(self, obj):
             many=True,
             context={'request': request}
         )
-
+        # Combining similar objects
```
emi-hi, line 262: is this part necessary?

vibhiquartech: Yes, it's combining the similar statuses to one & I have removed the comment.

backend/api/serializers/dashboard.py
```diff
@@ -258,12 +259,17 @@ def get_activity(self, obj):
             many=True,
             context={'request': request}
         )
-
+        
+        combined = defaultdict(int)
```
AlexZorkin, line 263: I don't think you need the merged_data here as long as the validation check for 'CHECKED' is in the serializer. Can you try removing this new section and see if the counts are correct?

AlexZorkin (changes requested):

backend/api/serializers/dashboard.py
vibhiquartech: When the validation status is RECOMMEND_APPROVAL, RECOMMEND_REJECTION, and CHECKED the status to be shown is SUBMITTED but the count is happening separately in the backend.
For example,  lets say we have 3  RECOMMEND_APPROVAL, 1 RECOMMEND_REJECTION  & 2 CHECKED, then the is backend sending the three SUBMITTED status separately thats why  I have combined the all three SUBMITTED status into one with the added count. If we remove that it will send multiple SUBMITTED status separately instead of one in the frontend.

AlexZorkin: Ok I see how it works now. Thanks.

AlexZorkin (approved):

AlexZorkin (approved):

PR #1360, opened 2023-02-03
Zeva 1306 - My Report status message bug fix

vibhiquartech:


backend/api/viewsets/model_year_report.py
```diff
@@ -450,7 +450,7 @@ def submission(self, request):
                 model_year_report_id=model_year_report_id,
                 validation_status=validation_status,
                 update_user=request.user.username,
-                create_user=request.user.username,
+                create_user=model_year_report_update[0].create_user,
```
emi-hi, line 453: This changes what is written to the database. It sets the create user to the user that created the initial model year report. We do not want this, because the initial creator of the model year report did not necessarily create the modelYearReportHistory record. If an analyst updates the status it will go in as the supplier having updated it, so the director would see history records that say something like 'supplier has recommended for approval'. 

 Not only does this result in the wrong data being added to that field, but it also does not address the issue in the card, as any older records are still going to show the original create_user. Are you intending to go through prod and update every single ModelYearReportHistory record?  Instead of changing what data is inserted into the database, change the way that data is displayed. We can control who sees what. We already are making sure that if it is in the recommended status, that bceid users only see that its submitted. Now we need to also check if its in the recommended status, that bceid users only see the original submitter. Idir users should still see the most accurate/up to date information. 

The card also outlines frontend changes that need to be made, which are not addressed here.

emi-hi (changes requested):

AlexZorkin (changes requested):
All of Emily's comments will need to be addressed. This branch needs to be rebased so you may want to create a new branch off of release first.

frontend/src/compliance/components/ComplianceReportAlert.js
```diff
@@ -36,6 +36,12 @@ const ComplianceReportAlert = (props) => {
           user: statusFilter('SUBMITTED').createUser.displayName
         };
       }
+      if (validationStatus === 'RECOMMENDED') {
```
emi-hi, line 39: This does not work either. The idea is to change what is passed on the backend so the frontend doesn't have to do any logic like this. See what we do with sales submissions for an example. We still want IDIR users to to see the most accurate and up to date information.

backend/api/viewsets/model_year_report.py
emi-hi: this needs to be taken out. We should not be changing how information is added to the database. Use the serializer to change what data is displayed.

emi-hi (changes requested):

emi-hi:
> All of Emily's comments will need to be addressed. This branch needs to be rebased so you may want to create a new branch off of release first.

My comments weren't addressed and it still needs to be rebased.

PR #1372, opened 2023-02-07
Front-End Unit Testing

JulianForeman:
Added front-end tests for Alert, Button, Comment and CommentInput components.  Also updated the CommentInput component to add id's to make testing it easier.


AlexZorkin (approved):
Looking great.

PR #1387, opened 2023-02-10
ZEVA-1353 - Drop heading from Model Year Supplementary Report/Reassessment

vibhiquartech:


frontend/src/supplementary/components/SupplementaryAnalystDetails.js
```diff
@@ -300,16 +300,18 @@ const SupplementaryAnalystDetails = (props) => {
...
         />
-      )}
-      <div className="row">
-        <div className="col">
-          <h2 className="mb-2 mt-3">
-            {isReassessment
-              ? `${reportYear} Model Year Report Reassessment`
-              : `${reportYear} Model Year Supplementary Report`}
-          </h2>
+        ) : (
```
emi-hi, line 303: the impression I got from the card was that the heading should be removed, so shouldn't it be removed from the code rather than conditionally rendered?

AlexZorkin (approved):

PR #1388, opened 2023-02-14
ZEVA-1367- Disable editing opening credit balances

vibhiquartech:


frontend/src/supplementary/components/CreditActivity.js
```diff
@@ -393,7 +393,7 @@ const CreditActivity = (props) => {
           newBalances={newBalances}
           newData={newData}
           pendingBalanceExist={false}
-          readOnly={!isEditable}
+          readOnly={isEditable}
```
AlexZorkin, line 396: This is confusing, readOnly should not be 'isEditable'

emi-hi, line 396: this is confusing

AlexZorkin (changes requested):
We don't want this field to be editable when assessed. This field should always only be calculated and never editable.

PR #1407, opened 2023-02-16
fix: zeva-1354 - a different solution

tim738745:
Looking at my original approach to zeva 1354, I wasn't too happy with it; I think this is better as the same information doesn't live in two states now.


AlexZorkin (approved):
Definitely cleaner. Nice update.

PR #1419, opened 2023-02-17
ZEVA-1367- Disable editing opening credit balances

vibhiquartech:


emi-hi (changes requested):
it looks like the opening balance is referenced 4 times in the code, twice (a and b) as you updated here, but theres also a separate case if no prior year balance exists, so theres an a and b there as well that needs to also be read only. Its just a little bit further down in the code (the input box starts at line 431 and the readonly line is 453 for A credits. B credits is after that.)

The other thing is in the meeting we just had with the PO, you agreed to make this plain text rather than a disabled input box. Please makes these changes for all 4 cases and then it can be merged in.

frontend/src/compliance/components/ComplianceObligationTableCreditsIssued.js
```diff
@@ -354,38 +345,17 @@ const ComplianceObligationTableCreditsIssued = (props) => {
...
-                          originalAValue: creditBalanceStart[each].A,
-                          originalBValue: creditBalanceStart[each].B,
-                          title: 'creditBalanceStart',
-                          modelYear: each,
-                          creditA: event.target.value
-                        });
-                      }}
-                      readOnly={readOnly}
-                    />
+                        ? formatNumeric(creditBalanceStart[each].A, 2)
```
AlexZorkin, line 348: This ternary is fairly redundant so let's use a method like the following instead:
const getSupplementalValue = (modelYear, creditAorBText) => {
  // creditAorB is either 'A' or 'B'
  const creditType = `credit${creditAorBText}Value`
  const newData = getNewData('creditBalanceStart', modelYear, creditType)
  if (newData == '' || newData == null) {
    formatNumeric(creditBalanceStart[modelYear][creditAorBText], 2)
  } else {
    return formatNumeric(newData)
  }
}

AlexZorkin (changes requested):
use the following method instead of the ternary operations:

This ternary is fairly redundant so let's use a method like the following instead:
const getSupplementalValue = (modelYear, creditAorBText) => {
  // creditAorB is either 'A' or 'B'
  const creditType = `credit${creditAorBText}Value`
  const newData = getNewData('creditBalanceStart', modelYear, creditType)
  if (newData == '' || newData == null) {
    formatNumeric(creditBalanceStart[modelYear][creditAorBText], 2)
  } else {
    return formatNumeric(newData)
  }
}

AlexZorkin (approved):

PR #1568, opened 2023-03-22
chore: zeva-1525 - remove references to the keycloak service account

tim738745:
@kuanfandevops this PR removes references to the keycloak service account from the deployment configs; it also modifies the nagios service (which I don't think we use anymore?) so that it no longer checks for a keycloak connection. Also removes the keycloak secret config. Can you please have a look to check if everything looks okay? Thanks!


kuanfandevops:
@tim738745 yes, we don't use Nagios any more, I'll remove entire Nagios

kuanfandevops (approved):
@tim738745 Looks good. I'll remove the reference on Openshift and also remove Nagios build.

Issue #1517, opened 2023-03-15
ZEVA - Incorrect credit balance being displayed for P. on organisation page


jdon-zev:
**Describe the Bug**
On the supplier index page in prod:
https://zeroemissionvehicles.gov.bc.ca/organizations

As per the screen shot below the credit balance of B credits for P is wrong. It is showing a deficit of [] B credits. Deficit are shown in brackets.

**Expected Behaviour**
The Vehicle Supplier index page should show the current and correct credit balance for all suppliers. For P their B credit balance should be zero (displayed as a dash). On the Supplier Credit Transactions list page the value is correctly being shown as 0. Investigate the discrepancy and fix the Vehicle Supplier index page to show the correct '0' value.

**Actual Behaviour**
The balance of B credits being displayed for P is incorrectly showing a deficit.

**Implications**
Not accurately displaying the number of credits held by suppliers undermines the credibility and usefulness of the ZEVA system.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role:
1. Go to https://zeroemissionvehicles.gov.bc.ca/organizations
2. Scroll down to P
3. See incorrect B deficit credit balance

vibhiquartech:
Based on my investigation here are my findings: 

- When the vehicle supplier does any credit transaction the transaction value gets inserted in the credit_transaction table in the DB and update the remaining balance in the account_balance table.
- The latest value for Credit A and Credit B on the Vehicle supplier Index page is fetched from the account_balance table. However, on the Supplier Credit Transactions list page, the values are fetched from the credit_transaction table and calculated in the front end before being displayed to the user. This leads to a discrepancy in the values displayed on the two pages.
- The backend calculation is error-free, and there are no validation errors in the front end. However, the balance might have been updated erroneously, leading to the current issue. Notably, this issue is only affecting the Porsche vehicle supplier and not any other supplier. To resolve the issue, updating the balance for Porsche for credit B to "0" could be a possible solution.

PR #1578, opened 2023-03-24
fix: ZEVA - Incorrect credit balance being displayed on organization page #1517

emi-hi:
fix: changes organization model method to use transactions to calculate balance rather than acount balance table
chore: removes unused serializer from compliance obligation (there are still some unused serializers here...
chore: removes unused references from some files
fix:adds check for array length in consumer sales ldv model table (was causing error if list length is 0)


AlexZorkin (approved):
This is great, it cleans up a lot of unused code!

PR #1622, opened 2023-04-03
chore: zeva-1499 - SupplementaryAnalystDetails unit tests

tim738745:
@AlexZorkin I'm not too sure about this test : // as an analyst
// if i view a supplemental report in submitted status
// I should be able to view a editable version of all 3 tabs

In the code, it looks like the "supplemental" and "reassessment" tabs should be non-editable when a supplemental report is in "submitted" status... I wrote the tests according to how they were described in the card, so the tests that deal with these 2 cases will fail.

### DIFF
--- a/frontend/src/supplementary/components/SupplementaryAnalystDetails.js
+++ b/frontend/src/supplementary/components/SupplementaryAnalystDetails.js
@@ -129,6 +129,7 @@ const SupplementaryAnalystDetails = (props) => {
           disabled={
             ['RECOMMENDED', 'ASSESSED'].indexOf(currentStatus) >= 0
           }
+          data-testid={'recommendation-' + each.id}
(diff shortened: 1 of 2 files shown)


tim738745:
> > @AlexZorkin I'm not too sure about this test : // as an analyst // if i view a supplemental report in submitted status // I should be able to view a editable version of all 3 tabs
> > In the code, it looks like the "supplemental" and "reassessment" tabs should be non-editable when a supplemental report is in "submitted" status... I wrote the tests according to how they were described in the card, so the tests that deal with these 2 cases will fail.
> 
> yah that is wrong oops! just the second tab should be editable, the first and third should be viewable but read only

Thanks @emi-hi ; I'll update the tests!

AlexZorkin (approved):

PR #1623, opened 2023-04-03
fix: zeva 1567 reassessment penalty showing up as $0

emi-hi:
fix: updates logic so that penalty is displayed for director and analyst in the written description for recommended/assessed reports as well as radio descriptions (for draft, returned, etc)
chore: removes formattedPenalty to reduce amount of code

### DIFF
--- a/frontend/src/supplementary/components/SupplementaryAnalystDetails.js
+++ b/frontend/src/supplementary/components/SupplementaryAnalystDetails.js
@@ -147,7 +147,7 @@ const SupplementaryAnalystDetails = (props) => {
                 details.assessmentData.legalName
               )
               .replace(/{modelYear}/g, details.assessmentData.modelYear)
-              .replace(/{penalty}/g, `$${formattedPenalty} CAD`)}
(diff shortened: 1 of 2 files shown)


frontend/src/supplementary/components/SupplementaryAnalystDetails.js
```diff
@@ -109,7 +105,7 @@ const SupplementaryAnalystDetails = (props) => {
           details.assessmentData.legalName
         )
         .replace(/{modelYear}/g, details.assessmentData.modelYear)
-        .replace(/{penalty}/g, `$${formattedPenalty} CAD`)
+        .replace(/{penalty}/g, `$${formatNumeric(supplementaryAssessmentData.supplementaryAssessment.assessmentPenalty, 0) || 0} CAD`)
```
AlexZorkin, line 108: we should add checks for null objects here just incase one in the chain breaks the view.

AlexZorkin (changes requested):

AlexZorkin (approved):

PR #1653, opened 2023-04-13
fix: ZEVA - Dashboard Loading Performance Issues for Director Role - 1612

emi-hi:
fix: moves setLoading(false) to the end of the getDashboard function, to happen after it finishes, adds a setLoading(true) to the beginning as well


JulianForeman (approved):
100% Faster than it was

AlexZorkin (approved):

PR #1660, opened 2023-04-14
feat: ldv sales from reassessments - 1636

emi-hi:
feat: adds beginning of service for adjusting credits based on reassessment, LDV sales from reassessments are now being written to the organizationLdvSales table and are displayed on the organization page and should be pulled into future model year reports

### DIFF
--- a/backend/api/serializers/model_year_report.py
+++ b/backend/api/serializers/model_year_report.py
@@ -268,8 +268,7 @@ class ModelYearReportListSerializer(ModelSerializer, EnumSupportSerializerMixin)
 
         def get_ldv_sales(obj):
             if (
-                consider_supplemental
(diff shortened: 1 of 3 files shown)


JulianForeman (changes requested):
1 Small thing, if you go to Vehicle Suppliers -> (Supplier with reassessment) -> Model Year Reports.  It uses the old data still.

Other than that everything is working!

JulianForeman:
- Removed consider_supplemental flag from get_ldv_sales function so that lists not containing the consider_supplemental? parameter will still grab reassessment data if the reassessment exists in an approved state and contains ldv sales data.

AlexZorkin (approved):

PR #1757, opened 2023-05-24
feat: zeva-1613 - credit request list sort by date and filter by enter

tim738745:


backend/api/services/sales_submission.py
```diff
@@ -0,0 +1,117 @@
...
+):
+    with connection.cursor() as cursor:
+        result = {}
+        sql_fields = ""
+        for index, value_field in enumerate(value_fields):
+            if index == 0:
+                sql_fields = sql_fields + value_field
+            else:
+                sql_fields = sql_fields + ", " + value_field
+        cursor.execute(
```
AlexZorkin, line 100: Are these parameter fields always coming from static strings within the backend? Just want to make sure there's no sql injection opportunities here.

AlexZorkin (approved):
Quite a complex problem to solve! nice work. Just the one sql injection question but otherwise looks good.

backend/api/services/sales_submission.py
tim738745: Currently, yes. But developers can pass anything they want to this function, including user input, so it's definitely something to be aware of.

PR #1826, opened 2023-06-29
Fix: Credit Transfer Signoff User

AlexZorkin:
fix: added check to exclude gov users from signoff history on credit transfers


kuanfandevops (approved):
looks good

Issue #1788, opened 2023-06-09
ZEVA - Credit Agreement/Adjustment modal confirmation


airinggov:
**Describe the Bug**
Not 100% sure this is a bug or something never implemented.
As an idir analyst or director when recommending or issuing credit agreements/adjustment (initiative agreement, purchase, agreement, administrative credit allocations/reductions, automatic administrative penalty) there are no modal confirmation pop-ups.

**Expected Behaviour**
A modal confirmation should appear to confirm primary actions. Modals should confirm the action being taken with a Cancel button or a button matching what they just clicked to reconfirm the action.

**Actual Behaviour**
No modal is appearing, the action occurs immediately upon first button click which doesn't match normal user experience.

**Implications**
There isn't a way to back out of an action which users have become accustomed to.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role: idir analyst & director
1. Go to a credit agreement
2. Click on a primary action button as an analyst or director
3. See no modal

PR #1827, opened 2023-06-29
Task - Credit Agreement Modal Confirmation #1788

JulianForeman:
- Added modal for confirming primary actions on credit agreements.

### DIFF
--- a/frontend/src/creditagreements/components/CreditAgreementsDetailsPage.js
+++ b/frontend/src/creditagreements/components/CreditAgreementsDetailsPage.js
@@ -38,7 +38,7 @@ const CreditAgreementsDetailsPage = (props) => {
       }}
       handleSubmit={() => {
         setShowModal(false)
-        handleSubmit((analystAction ? 'RECOMMEND' : 'ISSUE'))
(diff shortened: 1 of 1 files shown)


frontend/src/creditagreements/components/CreditAgreementsDetailsPage.js
```diff
@@ -25,6 +27,41 @@ const CreditAgreementsDetailsPage = (props) => {
...
+
+  const modal = (
+    <Modal
+      confirmLabel={(analystAction ? 'Submit' : 'Issue')}
+      handleCancel={() => {
+        setShowModal(false)
+      }}
+      handleSubmit={() => {
+        setShowModal(false)
+        handleSubmit((analystAction ? 'RECOMMEND' : 'ISSUE'))
```
tim738745, line 41: I think the statuses are supposed to be "RECOMMENDED" and "ISSUED"

tim738745 (changes requested):

frontend/src/creditagreements/components/CreditAgreementsDetailsPage.js
JulianForeman: Good catch, fixed.

Issue #1453, opened 2023-02-24
ZEVA - Index pages, React tables are missing the paging option


airinggov:
**Describe the Bug**
As an IDIR or BCeID Zeva user the main index pages with React tables are missing the paging component to set amount of records per page etc.

**BCeID**
Compliance Reporting
Credit Applications, Transfers
ZEV Models
Administration
•	Users

**IDIR**
Compliance Reporting
Credit Applications, Transfers, Agreement
ZEV Models
Vehicle Suppliers
•	Users
Administration
•	Users

**Expected Behaviour**
These index tables should have the paging component.

![paging.png](https://images.zenhubusercontent.com/63b7471574cb4a2d5b484c9f/7fd27076-cbff-4b6d-92cc-596fd6eab828)

**Actual Behaviour**
No paging component.

**Implications**
Users just get a long list that they can't adjust the amount of records displayed.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role: Any
1. Go to pages listed above
2. Scroll to bottom of React tables

**Acceptance Criteria**
- [ ] Given I am an IDIR user when I am viewing the vehicle supplier users, administration users, compliance reporting, credit applications, transfers, agreements, ZEV models or vehicle suppliers I should see the pages component to be able to set how many records I want to see on the page.
- [ ] Given I am a BCeID user when I am viewing the compliance reporting, credit applications, transfers, ZEV models or vehicle suppliers I should see the pages component to be able to set how many records I want to see on the page.
- [ ] Given I am a BCeID administrator when I am viewing the vehicle supplier users I should see the pages component to be able to set how many records I want to see on the page.

**Development Checklist**
- [ ] Add the paging component and have it display correctly on the pages listed.

**Additional Information**
Please also choose a table that is performing the most poorly (loads the most slowly) and implement server-side pagination for it - this will entail managing the state of the table (page number, number of pages, filters, sorts). (This is a different card - #1842)

kcabhar:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-Zelda-637d5bc12e6201001055b745/issues/bcgov/zeva/1453?planning-poker) with Zenhub @AlexZorkin @emi-hi @jig-patel @JulianForeman @tim738745 @vibhiquartech

PR #1856, opened 2023-07-14
Feat: Index Tables Pagination #1453

JulianForeman:
- Added front end pagination to all react tabes that require it.

### DIFF
--- a/frontend/src/app/components/ReactTable.js
+++ b/frontend/src/app/components/ReactTable.js
@@ -69,7 +69,20 @@ class CustomReactTable extends Component {
       showPagination
     } = this.props
 
-    return (
+    let paginationProps = {
+      pageSize: this.pageSize > 0 ? this.pageSize : 3,
+      pageSizeOptions: [this.pageSize],
+      showPagination: false,
+      defaultPageSize: this.defaultPageSize
+    }
+    
+    if (showPagination) {
+      paginationProps = {
+        showPagination: true,
+        defaultPageSize: 10
+      }
+    }    
+    return ( 
       <ReactTable
         ref={this.table}
         className={`searchable ${className}`}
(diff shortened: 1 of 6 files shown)


frontend/src/app/components/ReactTable.js
```diff
@@ -107,9 +108,7 @@ class CustomReactTable extends Component {
           this.pageSize =
             this.table.current.getResolvedState().sortedData.length
         }}
-        pageSize={this.pageSize > 0 ? this.pageSize : 3}
```
tim738745, line 110: Looks good, just wondering if we should preserve the option of using the custom table's original behaviour? Perhaps we could do something like:

let paginationProps = {
      //these are the original props
      pageSize: this.pageSize > 0 ? this.pageSize : 3,
      pageSizeOptions: [this.pageSize],
      showPagination: false,
      defaultPageSize: this.defaultPageSize
    }
    if (showPagination) {
      //these are the props we use if showPagination=true is passed
      paginationProps = {
        showPagination: true,
        defaultPageSize: 10
      }
    }

And then we pass in pagination props like so: {...paginationProps} to the component

tim738745 (changes requested):

frontend/src/app/components/ReactTable.js
JulianForeman: We can do this, but as you mentioned before the only other place that uses pagination uses the out of the box react-table for server side rendering and everywhere else now uses the custom one for strictly frontend.  What would you prefer to do?

tim738745: @JulianForeman let's preserve the option for the original behaviour, as I'm just concerned that someone may ask for a table that's just one big page

Issue #1868, opened 2023-07-25
ZEVA - report history


tim738745:
**Describe the Bug**
It looks like certain reassessments/supplementary reports are being excluded from the report history.

**Expected Behaviour**
I expect to see an accurate timeline of reports.

**Actual Behaviour**
There is a case where the most recent reassessment does not appear in the timeline.

**Implications**
This may confuse some users.

**Steps To Reproduce**
At the time of this writing: on dev, as a government user, go to Ford's 2021 reports. By **navigating** to the topmost (most recent) entry in the timeline, one will see a report that is not, in fact, the most recent report (which is supposed to be an assessed reassessment with a balance of 1598.88 A credits, and 0 B credits). Note that upon first landing on the page, you will see the correct report; it is only by navigating to a different entry then navigating back will one see a report that one should not be seeing.

**Notes**
I think this is a a backend issue; see the supplemental_history endpoint in viewsets/model_year_report.py. There is some logic in there concerning which supplemental reports to exclude that should probably be reviewed. Also, in SupplementalReportSerializer in serializers/model_year_report_noa.py, there is some logic about how to display the status of a supplemental report that should also be reviewed.

kcabhar:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-Zelda-637d5bc12e6201001055b745/issues/bcgov/zeva/1868?planning-poker) with Zenhub @AlexZorkin @JulianForeman @tim738745

PR #1920, opened 2023-08-14
Fix - Reassessment/Supplementary Report Timeline #1868

JulianForeman:
- Modified logic to exclude supplemental report from the reporting timeline if a reassessment has started for it

### DIFF
--- a/backend/api/viewsets/model_year_report.py
+++ b/backend/api/viewsets/model_year_report.py
@@ -339,29 +339,21 @@ class ModelYearReportViewset(
         exclude_supplemental_reports = []
         for report in supplemental_reports:
             if report.status.value in ["DRAFT", "RETURNED"] and (
-                request.user.is_government and not report.is_reassessment
-            ):
-                exclude_supplemental_reports.append(report.id)
+                    request.user.is_government and not report.is_reassessment
+                ):
+                    exclude_supplemental_reports.append(report.id)
 
             if report.is_reassessment:
                 if (
                     report.status.value not in ["ASSESSED", "REASSESSED"]
                     and not request.user.is_government
-                ):  
-                    exclude_supplemental_reports.append(report.id)
-
-                # If it's a reassessment report from a supplementary
-                # we need to remove the supplemental report from the history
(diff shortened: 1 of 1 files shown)


backend/api/viewsets/model_year_report.py
```diff
@@ -347,19 +347,21 @@ def supplemental_history(self, request, pk=None):
...
+                # in order to display the reassessment because it is the most recent report
+                # in the process
                 supplemental_report = SupplementalReport.objects.filter(
                     supplemental_id=report.supplemental_id
                 ).first()
-
+    
                 if supplemental_report and supplemental_report.from_supplemental:
-                    exclude_supplemental_reports.append(report.id)
+                    for re_report in supplemental_reports:
```
tim738745, line 362: I think the case we're trying to deal with here is simply when a supplier submits a supplementary, and the govt then recommends/assesses it. So that will create two records in the SupplementaryReports table, one with a status of submitted, and another with a status of recommended or assessed, and we want to exclude the submitted one. I think it would be cleaner to just do something like this inside the for loop (line 340):

`

           if report.status.value in ["DRAFT", "RETURNED"] and (
                request.user.is_government and not report.is_reassessment
            ):
                exclude_supplemental_reports.append(report.id)

            if report.is_reassessment:
                if (
                    report.status.value not in ["ASSESSED", "REASSESSED"]
                    and not request.user.is_government
                ):
                    exclude_supplemental_reports.append(report.id)
            else:
                if request.user.is_government and report.status.value == 'SUBMITTED':
                    next_report = SupplementalReport.objects.filter(supplemental_id=report.id).exclude(status=ModelYearReportStatuses.DELETED).first()
                    if next_report and next_report.is_reassessment:
                        exclude_supplemental_reports.append(report.id)`

tim738745 (changes requested):

backend/api/viewsets/model_year_report.py
JulianForeman: Didn't think about this solution, I implemented it.

tim738745 (approved):

Issue #1357, opened 2023-02-03
ZEVA - BUG - BCeID dashboard, broken link for Credit Agreements


airinggov:
**Describe the Bug**
The BCeID dashboard alert for Credit Agreements links to the Credit Agreements page which is only available on the IDIR side. The link can be updated to link to the main Credit Balance/Credit Transactions page as that is where they are displayed.
https://zeva-test.apps.silver.devops.gov.bc.ca/credit-transactions

**Expected Behaviour**
The BCeID dashboard alert for "Credit Agreements - # recorded by the Government of B.C." should link to
https://zeva-test.apps.silver.devops.gov.bc.ca/credit-transactions

no filters or parameters are required.

**Actual Behaviour**
Presently it links to
https://zeva-test.apps.silver.devops.gov.bc.ca/credit-agreements?col-status=Issued

which is wrong on the BCeID side 

**Implications**
A clear and concise description of any implications.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role:
1. Go to dashboard as BCEID
2. Click on credit agreement alert (need to have some recorded)

emi-hi:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-Zelda-637d5bc12e6201001055b745/issues/bcgov/zeva/1357?planning-poker) with Zenhub @tim738745 @AlexZorkin @JulianForeman @vibhiquartech @emi-hi

jdon-zev:
The link now re-directs the BCeID user to the credit transaction page but the page information does not load. All they see is a blank page.

shayjeff:
Credit agreement is opening to the credit transfers URL, when it should be opening to the credit transactions page.

PR #1968, opened 2023-09-28
Fix - Dashboard Link #1357

JulianForeman:
- Updated dashboard link to not filter under a specific circumstance and instead link to the general list


tim738745 (approved):

tim738745:
@JulianForeman looks good, just needs to be rebased onto the current release branch (1.53.0)!

Issue #1963, opened 2023-09-25
ZEVA - Initiative Agreement - different users shown for the same action


jdon-zev:
In production an initiative agreement has been recommended to the Director for them to issue the credits to the supplier. For the analyst role the correct user is shown as having made the action to recommend the agreement (see image Recommended by JADONALD) but for the Director role the screen shows a different user which is wrong (see image Recommended by CIANNICI).

In addition the correct users name should be shown, not their IDIR.  James Donald should be shown, not JADONALD, for example.

**Expected Behaviour**
When a user recommends an agreement to the Director their name should be shown in the page status banner. The name of the recommender should not change when the agreement is viewed by the Director or any other users.

**Actual Behaviour**
The analyst and the Director are seeing different IDIR's instead of the name of the person who has recommended the agreement to the Director in the status banner.

**Implications**
It is inaccurate and unclear to the Director who has recommended the agreement to them,

**Steps To Reproduce**
Steps to reproduce the behaviour: - (See attached screenshots)
User/Role:

1. Log in as a government analyst and Go to https://zeroemissionvehicles.gov.bc.ca/credit-agreements/2
2. Click on Recommend
3. Log in as a government director and Go to https://zeroemissionvehicles.gov.bc.ca/credit-agreements/2

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcjNVIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--c3d75012dd249a50b7c54c98eb58c7223abe561b/Recommended%20by%20JADONALD.jpg)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBci9VIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--021dda9ab4f8447c542b334f7368000042b9b37e/Recommended%20by%20CIANNICI.png)

**Dev Checklist**
- [ ] Update logic to use the user's first and last name instead of their username
- [ ] Change logic to not use the current user looking at the initiative agreement to whoever actually took the action
- [ ] Check to see if there is another function happening on this page (comment save maybe) that updates the recommended status or adds another one unintentionally.

kcabhar:
[...]

PR #1982, opened 2023-10-10
Fix: Initiative Agreement Different User Shown For Same Action #1963

JulianForeman:
- Updated serializer to include update user
- Front end user displayed now uses the update user for correct history

### DIFF
--- a/frontend/src/creditagreements/components/CreditAgreementsAlert.js
+++ b/frontend/src/creditagreements/components/CreditAgreementsAlert.js
@@ -30,13 +30,13 @@ const CreditAgreementsAlert = (props) => {
   switch (status) {
     case 'DRAFT':
       title = 'Draft'
-      message = `saved, ${date} by ${updateUser.firstName + " " + updateUser.lastName}.`
+      message = `saved, ${date} by ${typeof updateUser === 'string' ? updateUser : updateUser.firstName + " " + updateUser.lastName}.`
       classname = 'alert-warning'
       break
 
     case 'RECOMMENDED':
       title = 'Recommended'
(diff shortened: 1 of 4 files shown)


frontend/src/creditagreements/components/CreditAgreementsAlert.js
```diff
@@ -30,13 +30,13 @@ const CreditAgreementsAlert = (props) => {
   switch (status) {
     case 'DRAFT':
       title = 'Draft'
-      message = `saved, ${date} by ${user}.`
+      message = `saved, ${date} by ${updateUser.firstName + " " + updateUser.lastName}.`
```
tim738745, line 33: updateUser may be a string, so we should do a check: typeof updateUser === 'string' ? updateUser : updateUser.firstName + " " + updateUser.lastName

frontend/src/creditagreements/components/CreditAgreementsAlert.js
```diff
@@ -30,13 +30,13 @@ const CreditAgreementsAlert = (props) => {
...
       classname = 'alert-warning'
       break
 
     case 'RECOMMENDED':
       title = 'Recommended'
-      message = `recommended for issuance, ${date} by ${user}.`
+      message = `recommended for issuance, ${date} by ${updateUser.firstName + " " + updateUser.lastName}.`
```
tim738745, line 39: updateUser may be a string, so we should do a check: typeof updateUser === 'string' ? updateUser : updateUser.firstName + " " + updateUser.lastName

tim738745 (changes requested):

tim738745 (approved):

Issue #1939, opened 2023-08-30
ZEVA - Bug - Missing Comma


airinggov:
**Describe the Bug**
The LDV sales number displayed on both the Compliance Obligation and Assessment tabs is missing the comma.

**Expected Behaviour**
The LDV Sales number should have a comma as consistent throughout the site.

**Actual Behaviour**
The number is missing the comma

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBbXBWIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--4d54f309e7255d25d78849f16a213f12fa5bb9f0/missing-comma.png)

**Implications**
A clear and concise description of any implications.

AlexZorkin:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-Zelda-637d5bc12e6201001055b745/issues/bcgov/zeva/1939?planning-poker) with Zenhub @tim738745

jdon-zev:
ON dev all of these figures are still shown without the comma.

shayjeff:
Comma is still not appearing on the compliance obligation tab when in certain statuses.  It appears when it has been assessed but doesn't in other statuses.

JulianForeman:
Can you attach a screenshot of where you are still seeing the missing comma?

shayjeff:
[Uploading... image.png]

JulianForeman:
I can't see that image, could you try re-uploading it?

JulianForeman:
This issue should be resolved now but please test the same field while it's disabled and while it's editable because I had to update how we handle the input box specifically to accommodate a formatted string or a number respectively.

PR #2002, opened 2023-10-18
Fix: Final Missing Comma #1939

JulianForeman:
- Added logic to check for the input field being disabled if true the type it expects become a string and the value passed becomes a formatted number string instead of a number


tim738745 (approved):

tim738745 (changes requested):
Just need to address a couple of failing unit tests

tim738745 (approved):

Issue #1276, opened 2023-01-17
ZEVA - Return Model Year Report to Supplier


kcabhar:
**Description:**
When the analyst returns a MY report to the supplier any changes made by the analyst in the change log or comments should be cleared.

As an analyst when I start assessing a report I may make edits and add comments to the report. I may then decide to return the report to the supplier.

Presently when I do this, the supplier makes changes and send the report back to government, the report retains the original updates and comments made by the analyst. These should not appear now that the supplier has re-submitted the report.

When returning the report to the supplier it would be best to discard any changes the analyst has made as part of this process.

**Acceptance Criteria:**
Given I am an idir user
When I return a model year report to the supplier
Then any analyst edits, assessment radio selection, and penalty number should be 'deleted'. Comments between analyst/director can be retained.

**Development Checklist:**
add boolean fields to model_year_report_assessment_comment, model_year_report_assessment, model_year_report_make , model_year_report_ldv_sales
mark these records as display=false (or whatever the field is called) when the analyst returns to supplier
ensure that any viewsets using those tables does not display records that are display=false
update the changelog logic so it takes into account that there could be extra records rather than just 2 (ie filter out display=false before doing anything else) or preferably filter out on the backend

**Notes:**

kcabhar:
**James - 07/Sep/22**
Note - we decided to soft retain the previous information.

**Kimia- 12/Jan/23**
Following a discussion between James, Al and Kimia, this is card is to be done but in conjunction with a separate card to expand the report history to include a record of any report that is returned/sent back to a supplier.

shayjeff:
The LDV sales submitted by the analyst did not clear out. [Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBek95QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1d0044b1f52330cee1b7082a1cacd2998088369b/image.png)

shayjeff:
Everything is working as should except the comments between the analyst and the director are not being retained.  This is the only thing that should remain for the IDIR users. The PAC MY 2022 report shows the comments missing when the report was resubmitted by the supplier.

PR #2013, opened 2023-10-26
Feat: Returned Model Year Report Information #1276

JulianForeman:
- Adding display field to some report tables to allow reports to be returned, not show analyst adjustments and soft retain information


backend/api/viewsets/model_year_report.py
```diff
@@ -477,6 +477,56 @@ def submission(self, request):
...
+                display=False
+            )
+            ModelYearReportAssessmentComment.objects.filter(
+                model_year_report_id=model_year_report_id,
+                to_director=True
+            ).update(
+                display=False
+            )
+        
+        elif validation_status == "SUBMITTED":
```
tim738745, line 506: I'm not sure if we need to set display back equal to True; the card says "These should not appear now that the supplier has re-submitted the report." So I think we just set the display attributes of those objects to false when the report gets returned (set to Draft), and we don't reset them again? May need to confirm with Al/Shaylene...

JulianForeman: Oh yeah, I just interpreted it as we don't want the supplier to see analyst edits but the reason we were soft retaining the information was to have it show back up afterward. I will reach out to them for clarity.

JulianForeman: Reached out for more info and you are right; I removed the code that changes the display back to true.

Issue #1356, opened 2023-02-03
ZEVA - Make ZEV Models Active=Yes,- the default display 


airinggov:
**Title:** Make ZEV Models Active=Yes,- the default display on the ZEV Models page, both BCeID and IDIR

**Description:** Suppliers can make ZEV models inactive if they are no longer being sold. This keeps the ZEV Model page list more manageable (and also removes them from the credit application excel template). The filter on the Action Yes/No column should default to Yes. Even when the Clear Filters button is clicked it should keep the Yes filter. Users can click the Show Inactive ZEV button to change the filter to No (and of course they can also type Yes or No).

The dash is also required to see models that are in Draft, Submitted and Changes Requested statuses

This change should be done for all BCeID and IDIR users/roles.

**Wireframe:** https://preview.uxpin.com/4946124c44da522837a3b54ccb7b5243fa384039#/pages/131174642?mode=fm

**Purpose and benefit to user:** It keeps the page display more manageable as was intended by this feature.

**Acceptance Criteria:**

- [ ] Given I am a BCeID or IDIR user, When I am looking at the ZEV Models List page, then the default view filters out Inactive ZEV.

**Development Checklist:**
- [ ] Update so when the VehicleListTable displays, it always defaults to filtering out the inactive models when the page loads
- [ ] The only time the inactive models should appear is if 'show inactive models' button is clicked or if 'no' is typed in the column filter
- [ ] When 'clear filters' button is pressed, reset all filters except set the active filter to not show inactive models. If they want to clear all filters except the active/inactive models, they will need to manually type no in that column

**Notes:**

emi-hi:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-Zelda-637d5bc12e6201001055b745/issues/bcgov/zeva/1356?planning-poker) with Zenhub @AlexZorkin @emi-hi @JulianForeman @tim738745 @vibhiquartech

airinggov:
I updated this card to add information about the default filter needs to include a dash (Yes,-) so the points of the card may need altered from 3

airinggov:
The filter isn't including the -/dash. This is an issue as it doesn't display ZEV in draft, submitted, rejected status that are neither Active Yes or No. Filtering used to support multiple values separated by a comma but it doesn't appear to work now.

PR #2014, opened 2023-10-26
Feat: Active ZEV Models Default Filter #1356

JulianForeman:
- Modifying the filtering on the ZEV models list page to default to only active models

### DIFF
--- a/frontend/src/vehicles/VehicleListContainer.js
+++ b/frontend/src/vehicles/VehicleListContainer.js
@@ -23,12 +23,6 @@ const VehicleListContainer = (props) => {
   const query = qs.parse(location.search, { ignoreQueryPrefix: true })
   const handleClear = () => {
     setFiltered([{id: 'is-active', value: 'Yes'}])
-
-    props.history.push({
-      pathname: location.pathname,
-      search: '', 
-      state: {}
-    });
   
     refreshList(true);
   }
--- a/frontend/src/vehicles/VehicleListContainer.js
+++ b/frontend/src/vehicles/VehicleListContainer.js
@@ -13,7 +13,7 @@ import VehicleList from './components/VehicleList'
 const qs = require('qs')
 
 const VehicleListContainer = (props) => {
-  const [filtered, setFiltered] = useState([])
+  const [filtered, setFiltered] = useState([{id: 'is-active', value: 'Yes'}])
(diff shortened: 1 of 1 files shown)


frontend/src/vehicles/VehicleListContainer.js
```diff
@@ -22,27 +22,45 @@ const VehicleListContainer = (props) => {
 
   const query = qs.parse(location.search, { ignoreQueryPrefix: true })
   const handleClear = () => {
-    setFiltered([])
+    setFiltered([{id: 'is-active', value: 'Yes'}])
+
+    props.history.push({
```
tim738745, line 27: Just curious why this was added?

frontend/src/vehicles/VehicleListContainer.js
```diff
@@ -22,27 +22,45 @@ const VehicleListContainer = (props) => {
...
+      pathname: location.pathname,
+      search: '', 
+      state: {}
+    });
+  
+    refreshList(true);
```
tim738745, line 33: Since the filtering is frontend-only for this component, do we need to refresh when clearing the filters?

JulianForeman: I found that when it wasn't refreshed the fields of the table don't get reset and the filters don't either.

frontend/src/vehicles/VehicleListContainer.js
JulianForeman: I was having weird issues with clearing the filters and just wanted to make sure the browser wasn't caching anything. I removed it

frontend/src/vehicles/VehicleListContainer.js
tim738745: I tried it out; it looks like there's an issue where if there are filters present and you clear the filters, the shouldComponentUpdate() method of the custom react table we use gives us an error. So yes, refreshing is probably the easiest thing to do here. In that case, we can remove the "setFiltered([{id: 'is-active', value: 'Yes'}])" from the handleClear() function since refreshList() inserts that filter into the filters list.

JulianForeman: I forgot I doubled up on that filter in the refreshList function. I removed the unnecessary filtering bit.

tim738745 (approved):

PR #2028, opened 2023-10-31
Feat: Returned Model Year Report Information #1276

JulianForeman:
- Adding display field to some report tables to allow reports to be returned, not show analyst adjustments and soft retain information


backend/api/viewsets/model_year_report.py
```diff
@@ -477,6 +477,31 @@ def submission(self, request):
...
+                from_gov=True
+            ).update(
+                display=False
+            )
+            ModelYearReportAssessment.objects.filter(
+                model_year_report_id=model_year_report_id
+            ).update(
+                display=False
+            )
+            ModelYearReportAssessmentComment.objects.filter(
```
tim738745, line 499: display is set to 'false' here for ModelYearAssessmentComment, but the display field is never read, meaning that analysts would still see old assessment comments after re-submission?  Perhaps we should confirm this and if this is the case, then we could address this is in the frontend (AssessmentContainer.js)

tim738745 (changes requested):

backend/api/viewsets/model_year_report.py
JulianForeman: Confirmed the logic, and I added a bit to not display the assessment comments if the report is returned and then re-submitted.

backend/api/migrations/0004_auto_20231024_0908.py
```diff
@@ -0,0 +1,464 @@
...
+            model_name='modelyearreportldvsales',
+            name='display',
+            field=models.BooleanField(default=True),
+        ),
+        migrations.AddField(
+            model_name='modelyearreportmake',
+            name='display',
+            field=models.BooleanField(default=True),
+        ),
+        migrations.AlterField(
```
tim738745, line 34: Looks good, just one last thing: I noticed the backend unit tests failing, and I tried it out and realized that it's because of some of these "AlterField" migrations. These migrations are being created because we upgraded the Django version a few releases ago, and the primary key + foreign key fields of our tables are being changed from AutoField to BigAutoField. This is fine and shouldn't be a breaking change, but I think what's happening is we have a legacy "AccountBalance" model, and because of a setting on the model, its foreign keys can't be updated to type BigAutoField. What I suggest we do right now is we remove the "AlterField" migrations and just keep the "AddField" migrations that are a result of the changes you made to the models (so remove lines 34 to 463). We can address these AlterField migrations later, and probably fix the current problem by removing the AccountBalance table entirely.

tim738745 (changes requested):

tim738745 (approved):

PR #2050, opened 2023-11-08
Fix: Report Returned Fixes #1276

### LINKED ISSUE #1276: ZEVA - Return Model Year Report to Supplier

JulianForeman:
- Added logic to not display assessment comments, radio button selections or penalty amounts when a report is returned

### DIFF
--- a/backend/api/serializers/model_year_report.py
+++ b/backend/api/serializers/model_year_report.py
@@ -115,7 +115,7 @@ class ModelYearReportSerializer(ModelSerializer):
         ) or request.user.is_government
 
         if is_assessed:
-            return obj.get_ldv_sales() or obj.ldv_sales()
+            return obj.get_ldv_sales(from_gov=True) or obj.ldv_sales
 
         return obj.ldv_sales
 
--- a/backend/api/serializers/model_year_report.py
+++ b/backend/api/serializers/model_year_report.py
@@ -296,7 +296,8 @@ class ModelYearReportListSerializer(ModelSerializer, EnumSupportSerializerMixin)
                     return "-"
 
                 assessment = ModelYearReportAssessment.objects.filter(
-                    model_year_report_id=obj.id
+                    model_year_report_id=obj.id,
+                    display=True
                 ).first()
 
                 if assessment:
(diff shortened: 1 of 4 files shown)


tim738745 (approved):

backend/api/serializers/model_year_report.py
```diff
@@ -108,21 +109,20 @@ def get_confirmations(self, obj):
...
         request = self.context.get("request")
-
         is_assessed = (
             request.user.organization_id == obj.organization_id
             and obj.validation_status == ModelYearReportStatuses.ASSESSED
         ) or request.user.is_government
 
         if is_assessed:
-            return obj.get_ldv_sales(from_gov=True) or obj.ldv_sales
+            return obj.get_ldv_sales() or obj.ldv_sales()
```
tim738745, line 118: This change is causing a unit test to fail: 
File "/github/workspace/api/tests/test_model_year_reports.py", line 71, in test_assessment_patch_response ... 
File "/github/workspace/api/serializers/model_year_report.py", line 118, in get_ldv_sales
    return obj.get_ldv_sales() or obj.ldv_sales()
TypeError: 'NoneType' object is not callable

So either the unit test has to be updated or we have to reconsider this change

tim738745 (changes requested):

backend/api/serializers/model_year_report.py
JulianForeman: I undid this change so the test should pass.

tim738745 (approved):

tim738745:
@JulianForeman just needs to be rebased into 1.55.0

Issue #2063, opened 2023-11-24
ZEVA - Internal commenting disappearing on credit agreements 


airinggov:
**Title:** Internal commenting disappearing on credit agreements when issued

**Description:** The comments between the analyst and director disappear after a credit agreement is issued.

**Wireframe:** <link>
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNlVIQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--10d6e04b803422184ad1f1bd4a54e0604e534ed4/image.png)

**Purpose and benefit to user:** The internal comments should remain visible after the transaction has been issued. Only on the IDIR side.

**Acceptance Criteria:**

- [ ] Given I am a IDIR user, When I am looking at an issued credit agreement, then I can see any comments that were made between the analyst and director

**Development Checklist:**

- [ ] I think there's just a conditional in a display component that makes it so that internal comments are not displayed when the status is "Issued" (or "Validated", whatever the final status is). So perhaps the only thing to do is remove/change this conditional.

**Notes:**

kcabhar:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-Zelda-637d5bc12e6201001055b745/issues/bcgov/zeva/2063?planning-poker) with Zenhub @tim738745 @emi-hi

PR #2069, opened 2023-12-06
Feat: Credit Agreement Internal Commenting #2063

JulianForeman:
- Allowing comments to appear if a credit agreement is issued without the option to add more comments


JulianForeman:
Outdated

PR #2089, opened 2023-12-11
Feat: zeva 2061: MY report history wording refinement

emi-hi:
feat: adds more information to history items with returned status, so idir users can tell who the report was returned by

### DIFF
--- a/frontend/src/compliance/components/ComplianceHistory.js
+++ b/frontend/src/compliance/components/ComplianceHistory.js
@@ -66,16 +66,31 @@ const ComplianceHistory = (props) => {
     if (itemHistory) {
       itemHistory.forEach((obj, i) => {
         if (['DRAFT'].indexOf(obj.status) >= 0) {
-          if ((itemHistory[i + 1]?.status === 'SUBMITTED' || itemHistory[i + 1]?.status === 'RETURNED') && (!item.isSupplementary || (itemHistory.supplementalReportId === itemHistory[i + 1].supplementalReportId))) {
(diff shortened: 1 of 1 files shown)


frontend/src/compliance/components/ComplianceHistory.js
```diff
@@ -66,16 +66,31 @@ const ComplianceHistory = (props) => {
...
+            // if the status is draft and the previous status was submitted or returned,
+            // it was returned to the supplier by the analyst
             const actuallyReturned = { ...obj }
-            actuallyReturned.status = 'RETURNED'
+            actuallyReturned.status = 'RETURNED TO THE SUPPLIER'
             tempHistory.push(actuallyReturned)
-          } else if ((itemHistory[i + 1]?.status !== 'SUBMITTED' && itemHistory[i + 1]?.status !== 'RETURNED') || user.isGovernment) {
+          } 
+          else if ((itemHistory[i + 1]?.status !== 'SUBMITTED'
+            && itemHistory[i + 1]?.status !== 'RETURNED'))  {
```
tim738745, line 78: I think we need to keep the "|| user.isGovernment" part in this conditional. I think this is for the case where a supplier submits a supplementary, and the analyst creates a reassessment from that supplementary. Then the history would look like ["draft", "submitted"] (from most recent to earlier) and we want to display this "draft" status as a "created" history item if the user is a government analyst.

tim738745 (changes requested):

tim738745 (approved):

PR #2094, opened 2023-12-15
fix ZEVA 2091: credit application table filter missing some submitted records

emi-hi:
adds recommended and checked submissions to the queryset filter if the bceid user searches anything that matches submitted

### DIFF
--- a/backend/api/viewsets/credit_request.py
+++ b/backend/api/viewsets/credit_request.py
@@ -228,10 +228,8 @@ class CreditRequestViewset(
                 if submitted_search == True:
                     # if the user has used anything that matches submitted on the frontend
                     # ie 'sub' or 'submit' it will return checked or recommended too
-                    connector = final_q.connector
-                    final_q.add(('validation_status__exact', 'CHECKED'), connector)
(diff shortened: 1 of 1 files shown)


backend/api/viewsets/credit_request.py
```diff
@@ -214,6 +214,22 @@ def filter_by_status(self, queryset, search_type, search_terms):
         )
 
         if final_q:
+            if not user.is_government:
```
tim738745, line 217: A cleaner way to do this might be to conditionally add the statuses to mapped_search_terms before the final_q is constructed; for example, on line 208, you could do:

if not user.is_government and SalesSubmissionStatuses.SUBMITTED.value in mapped_search_terms:
            mapped_search_terms.append(SalesSubmissionStatuses.CHECKED.value)
            mapped_search_terms.append(SalesSubmissionStatuses.RECOMMEND_APPROVAL.value)
            mapped_search_terms.append(SalesSubmissionStatuses.RECOMMEND_REJECTION.value)

tim738745 (changes requested):

Issue #1983, opened 2023-10-11
ZEVA - Credit Application Back Button


shayjeff:
**Describe the Bug**
As an IDIR analyst when I access the credit application under the vehicle supplier tab then hit the back button it loads a incomplete page (URL loses organization ID).
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1FwQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--55513d4cc3b5d2634ce1041375bfdceb612b6838/image.png)

**Expected Behaviour**
Return to the same screen that I started from. 

**Actual Behaviour**
It loses the organization ID and can't load the page properly. 

**Implications**
Breaks the navigation flow.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role: IDIR User
1. Click on vehicle supplier tab 
2. Select a supplier
3. Go to credit transactions sub tab
4. Select a credit application from the transactions 
5. Hit the back button 
6. See error

**Dev Checklist**
- [ ] Fix routing of back button credit applications to send back to the list of credit applications

kcabhar:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-Zelda-637d5bc12e6201001055b745/issues/bcgov/zeva/1983?planning-poker) with Zenhub @JulianForeman @tim738745

PR #2115, opened 2024-01-17
Fix: Credit Application Back Button #1983

JulianForeman:
- Updated back button routing to return user to the vehicle suppliers credit transactions list or all supplier transactions list based on href state


tim738745 (approved):

tim738745:
@JulianForeman Looks good, just needs to be rebased onto 1.58.0!

Issue #2125, opened 2024-01-23
ZEVA - Display timestamps in Pacific Time


tim738745:
**Title**
Display timestamps in Pacific Time

**Description:**
Display timestamps in Pacific Time

**Wireframe:** <link>

**Purpose and benefit to user:**

**Acceptance Criteria:**

- [ ] Given I am a user, When I am viewing a timestamp, then it's in Pacific Time

**Development Checklist:**
- [ ] Use setDefault() of moment.js to set the default timezone (https://momentjs.com/timezone/docs/#/using-timezones/default-timezone/)

**Notes:**

shayjeff:
Testing completed: Collaborated with the Sr. Developer as this was challenging to replicate. The Sr. Dev modified the code so I could observe both the incorrect timestamp and the updated timestamp during testing.

PR #2128, opened 2024-01-30
Feat: Timezones Default PST #2125

JulianForeman:
- Setting the default timezone to PST at the top of the application to display timestamps correctly.

### DIFF
--- a/frontend/src/app/app.js
+++ b/frontend/src/app/app.js
@@ -11,8 +11,6 @@ import moment from 'moment-timezone'
 import 'toastr/build/toastr.min.css'
 import 'react-table/react-table.css'
 
-moment.tz.setDefault("America/Los_Angeles")
-
 class App extends Component {
   constructor (props) {
     super(props)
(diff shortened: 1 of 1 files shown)


frontend/src/app/app.js
```diff
@@ -6,10 +6,13 @@ import Loading from './components/Loading'
 import CONFIG from './config'
 import Login from './Login'
 import Router from './router'
+import moment from 'moment-timezone'
 
 import 'toastr/build/toastr.min.css'
 import 'react-table/react-table.css'
 
+moment.tz.setDefault("America/Los_Angeles")
```
tim738745, line 14: Can we change this to "America/Vancouver" (just in case) and can we move this line into the constructor below? Like:
```
constructor (props) {
    super(props)
    this.state = {
      authenticated: false,
      keycloak: null
    }
    this.logout = this.logout.bind(this)
    moment.tz.setDefault("America/Los_Angeles")
  }
```

tim738745 (changes requested):

tim738745 (approved):

PR #2134, opened 2024-02-09
Task: BigAutoField Migrations Fix #2046

### LINKED ISSUE #2046: ZEVA - BigAutoField migrations
**Describe the task**
When we updated Django (to 3.2.X), DEFAULT_AUTO_FIELD was changed to BigAutoField. This means that every time we makemigrations, AlterField operations will be created to change the primary keys and associated foreign keys of our tables from type integer to BigInteger. 

If we accept these operations, the django testing framework can't set up the test database to run our tests because of an issue with our AccountBalance model (the issue is probably the fact that AccountBalance is not managed).

We should either fix the AccountBalance issue and do these operations or figure out a way to disable them from 
appearing each time we makemigrations.

**Purpose**
We need to be able to accommodate this Django change.

**Acceptance Criteria**
- [ ] Read https://docs.djangoproject.com/en/4.2/releases/3.2/#customizing-type-of-auto-created-primary-keys and see if we can prevent these operations from appearing.
- [ ] If the above doesn't work, check whether deleting the AccountBalance model (which we no longer use) fixes the issue with the tests, and if so, check the production db to see if we still need the account_balance table there, and if not, go ahead and delete the AccountBalance model.

**Additional context**

JulianForeman:
- Updating settings.py to change default field back to older version

### DIFF
--- a/backend/zeva/settings.py
+++ b/backend/zeva/settings.py
@@ -41,7 +41,7 @@ DEBUG = os.getenv('DJANGO_DEBUG', 'False') == 'true'
 DEVELOPMENT = os.getenv('DEVELOPMENT', 'False') == 'True'
 TESTING = 'test' in sys.argv
 RUNSERVER = 'runserver' in sys.argv
-DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
+DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
 
 ALLOWED_HOSTS = ['*']
 


tim738745 (approved):
This should work; I noticed that even with setting DEFAULT_AUTO_FIELD to AutoField, some migrations are still being created updating Foreign Key fields to type int, even though they are already of type int. So this shouldn't be an issue, but just something to be aware of.

PR #2135, opened 2024-02-12
feat: zeva-2112 - refined emails

tim738745:


tim738745:
@kuanfandevops a new environment variable called FRONTEND_URL_BASE needs to be added; value should be "https://zeva-dev.apps.silver.devops.gov.bc.ca" in dev, "https://zeva-test.apps.silver.devops.gov.bc.ca" in test, and "https://zeroemissionvehicles.gov.bc.ca" in prod. Thanks!

PR #2148, opened 2024-04-04
fix: 2146 - compliance ration reduction calculation

tim738745:


tim738745:
Same code as in https://github.com/bcgov/zeva/pull/2147, which was approved

PR #2217, opened 2024-08-16
Feat 2206 - Prevent commentator's name from displaying when updating information from comment

emi-hi:
removes commentors name from vehicle and credit requests

if commenter is idir and current user is bceid:
show 'government user'

if commenter is bcied or current user is idir:
show details

### DIFF
--- a/backend/api/serializers/sales_submission_comment.py
+++ b/backend/api/serializers/sales_submission_comment.py
@@ -7,7 +7,6 @@ from api.models.user_profile import UserProfile
 from api.models.sales_submission_comment import SalesSubmissionComment
 from api.serializers.user import MemberSerializer
 
-
 class SalesSubmissionCommentSerializer(ModelSerializer):
     """
     Serializer for sales submission comments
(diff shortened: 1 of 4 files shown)


backend/api/serializers/sales_submission_comment.py
```diff
@@ -7,20 +7,26 @@
...
-        return serializer.data
+        if not commenting_user.is_government or request.user.is_government:
+            ## if the commentor is not government or the request
+            #  user is government, show all the data
+            serializer = MemberSerializer(commenting_user, read_only=True)
+            return serializer.data
+        else:
+            #if the request user is not government and the commenter
+            #is government
+            return {'display_name': 'Government User'}
```
tim738745, line 29: I see that this `get_create_user()` method used to return the data associated with a `MemberSerializer` instance, which consists of fields `('id', 'first_name', 'last_name', 'email', 'display_name', 'is_active', 'phone', 'roles', 'is_mapped')`. Just wanted to make sure that returning only a `display_name` doesn't break anything in the frontend (e.g. if `email` is null it doesn't throw an error). As long as that's the case, looks good to me!

backend/api/serializers/vehicle_comment.py
```diff
@@ -15,12 +15,19 @@ class VehicleCommentSerializer(ModelSerializer):
...
-        return serializer.data
+        if not commenting_user.is_government or request.user.is_government:
+            ## if the commentor is not government or the request
+            #  user is government, show all the data
+            serializer = MemberSerializer(commenting_user, read_only=True)
+            return serializer.data
+        else:
+            #if the request user is not government and the commenter
+            #is government
+            return {'display_name': 'Government User'}
```
tim738745, line 30: Same comment as above

backend/api/serializers/sales_submission_comment.py
tim738745: ok, sounds good, thanks for checking! I'll merge it in

Issue #2222, opened 2024-08-20
ZEVA - forecast report confirmation checkbox


tim738745:
Please implement the confirmation checkbox associated with the forecast report (please see https://www.figma.com/design/ijCyqDXXGVGDmmyfaXzPhk/ZEVA?node-id=50-193). The save button should be disabled until this checkbox and all other checkboxes are checked. After saving, if the user clicks the "Edit" button (top right of page), then all the checkboxes should become unchecked.

katerinkus:
Both situations worked for me.

PR #2232, opened 2024-08-28
Task: Sales Forecast Report Checkbox Assertion #2222

JulianForeman:
- Adding new signing authority assertion for Sales Forecast in database.
- Updated frontend state logic to accomodate 2 checkboxes.


frontend/src/compliance/components/ConsumerSalesDetailsPage.js
```diff
@@ -208,13 +215,14 @@ const ConsumerSalesDetailsPage = (props) => {
...
             <div className="col-12 my-3">
-              <ComplianceReportSignOff
-                assertions={assertions}
-                checkboxes={checkboxes}
-                handleCheckboxClick={handleCheckboxClick}
-                disabledCheckboxes={disabledCheckboxes}
-                user={user}
-              />
+            <ComplianceReportSignOff
+              assertions={assertions.filter(assertion => assertion.displayOrder === 5)}
```
tim738745, line 219: We can just use this component for all assertions (please see comment above), so we can revert this change.

frontend/src/compliance/components/ConsumerSalesDetailsPage.js
```diff
@@ -87,14 +88,8 @@ const ConsumerSalesDetailsPage = (props) => {
...
 
-  assertions.forEach((assertion) => {
-    if (checkboxes.indexOf(assertion.id) >= 0) {
-      disabledCheckboxes = 'disabled'
-    }
-  })
-
   const disableSave = () => {
-    if (vehicles.length <= 0 && !checked) {
+    if (checkboxes.length >= 2 && vehicles.length <= 0) {
```
tim738745, line 92: I think `checkboxes` is a list of ids of assertions that have been checked, so what makes more sense to me is `checkboxes.length !== assertions.length`, and I don't think we need the `vehicles.length <= 0` part.

frontend/src/compliance/ConsumerSalesContainer.js
```diff
@@ -122,21 +129,24 @@ const ConsumerSalesContainer = (props) => {
...
-      forecastRecords: forecastRecords,
-      ...forecastTotals
-    })
-    Promise.all([consumerSalesPromise, forecastPromise]).then(() => {
-      history.push(ROUTES_COMPLIANCE.REPORTS)
-      history.replace(
-        ROUTES_COMPLIANCE.REPORT_CONSUMER_SALES.replace(/:id/g, id)
-      )
-    })
+    if(checkboxes.length >= 2){
```
tim738745, line 132: I think checkboxes is a list of ids of assertions that have been checked, so what makes more sense to me is `checkboxes.length === assertions.length`

tim738745 (changes requested):

tim738745 (approved):

PR #2234, opened 2024-08-29
ZEVA 2216: Fix - Rounding Error Between Credit Transaction Line Items and Detailed Credit Balance

emi-hi:
work done by @tim738745


tim738745:
@emi-hi, I'm actually a bit wary of this change because now `getModelYearReportCreditBalances()` returns an object containing Bigs instead of numbers, and other functions that rely on its output may not know how to deal with that.

Can we test this a bit more? For example, if a supplier is in deficit, what happens? Also, does everything work as expected on the BCeID side as well (e.g.g does the balance at the top right corner render correctly?)

tim738745 (approved):

Issue #2223, opened 2024-08-20
ZEVA - conditional rendering of "Vehicles Supplied" wording and update "Vehicle Class"


tim738745:
This is related to card #2205  - the "Vehicles Supplied" wording should only apply to MYRs associated with the 2024 compliance year or later (earlier MYRs should continue to have the original wording).

**Describe the task**
Remove "vehicle class" visibility for BCeID users (suppliers) while keeping it visible for IDIR users. On the supplier side, flag any vehicle entries with a weight over 4536 kg. Additionally, hide the classification "2B" if the vehicle weight is between 3856 kg and 4537 kg.

**Purpose**
This will eliminate the use of a vehicle class that is currently not referenced in the Act or Regulations and ensure proper handling of vehicle weight information.

**Acceptance Criteria**
- [ ] Vehicle classification is hidden for suppliers but visible to Idir users.
- [ ] Any vehicle entry over 4536 kg is flagged on the supplier side.
- [ ] Classification "2B" is hidden if the vehicle weight is between 3856 kg and 4537 kg.

Additional context
•	This was new information from the BA after card 2203 was completed. Tim asked me to add the detail to this card. If you have any questions please let us know.

katerinkus:
- [x] Vehicle classification is hidden for suppliers but visible to Idir users.
- [ ] Any vehicle entry over 4536 kg is flagged on the supplier side.
- [x] Classification "2B" is hidden if the vehicle weight is between 3856 kg and 4537 kg.

PR #2235, opened 2024-08-29
Task: Conditionally Render Vehicle Weight Class Text #2223

JulianForeman:
- Conditionally render text for the vehicle class if it is an idir user. 
- Removes 2B flag for everyone.
- Conditional rendering for vehicles supplied text depending on model year.

### DIFF
--- a/frontend/src/app/components/TextInput.js
+++ b/frontend/src/app/components/TextInput.js
@@ -6,7 +6,6 @@ const TextInput = (props) => {
     additionalClasses,
     defaultValue,
     errorMessage,
-    helperText,
     handleInputChange,
     label,
     name,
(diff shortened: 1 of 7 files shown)


tim738745:
Hi @JulianForeman, I think there was some confusion regarding this card - the original card I wrote asked for:

(1) the conditional rendering of the new "Vehicles Supplied" text, 

and then Shaylene added a new task to the card that asked for:

(2) the "2B" designation to be hidden from suppliers. 

So I see you did (2), but missed (1). For (1), it just means going back to your PRs 2212 and 2213 and looking at where the "Vehicles Supplied" text was added, and making sure that the new text is rendered only for MYRs of compliance year 2024 or later. So, for example, in `ComplianceObligationAmountsTable`, instead of just changing the `{reportYear} Model Year LDV Sales:` line to `{reportYear} Vehicles Supplied:`, we would do something like:

`{reportYear} {reportYear < 2024 ? "Model Year LDV Sales" : "Vehicles Supplied"}:`

Thanks and sorry for the confusion!

tim738745 (approved):

PR #2285, opened 2024-10-02
2256 Task: Update Compliance Ratios

emi-hi:
task: adds fixtures for updating compliance ratios and deleting unnecessary ones


backend/api/fixtures/operational/0037_update_compliance_ratio.py
```diff
@@ -0,0 +1,50 @@
+from django.db import transaction
+from api.management.data_script import OperationalDataScript
+from api.models.compliance_ratio import ComplianceRatio
+
+class UpdateComplianceRatio(OperationalDataScript):
```
tim738745, line 5: Instead of doing this custom fixture, can we do a data migration instead (https://docs.djangoproject.com/en/5.1/topics/migrations/#data-migrations)?

It just seems like that's the standard way to update data in Django; you can also look at this PR Julian did for reference: https://github.com/bcgov/zeva/pull/2232/files (specifically the 0008 migration)

tim738745 (changes requested):

tim738745 (approved):

PR #2304, opened 2024-10-11
ZEVA feat 2284 - reduction

emi-hi:
feat- removes code that automatically reduces transaction credits if deficit exists


tim738745 (approved):

tim738745:
@emi-hi looks good to me! Just needs to be rebased onto release-1.63.0!

PR #2320, opened 2024-10-21
Chore/2267 - remove commentator's name

emi-hi:
creates user mixin for displaying update or create users as a specific way depending on whether they are government or not and whether the request user is government or not

refactors code to use this mixin for both commenters and other update/create users

### DIFF
--- a/backend/api/mixins/user_mixin.py
+++ b/backend/api/mixins/user_mixin.py
@@ -1,3 +1,4 @@
+from rest_framework.serializers import ModelSerializer, SerializerMethodField
 from api.models.user_profile import UserProfile
 from api.serializers.user import UserBasicSerializer
 from api.serializers.organization import OrganizationNameSerializer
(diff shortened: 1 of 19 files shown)


backend/api/mixins/user_mixin.py
```diff
@@ -0,0 +1,28 @@
+from rest_framework import serializers
+from ..models.user_profile import UserProfile
+from ..serializers.user import UserBasicSerializer
+from ..serializers.organization import OrganizationNameSerializer
+class UserMixin:
```
tim738745, line 5: From what I can tell, you're using this mixin as a superclass for serializers, and also as something to be instantiated so that you can invoke the get_user_data() method (in `services/model_year_report.py`). Since that's the case, I think something like this might be a bit cleaner:

```
def get_user_data(obj, user_attr, request):
    # your logic here

class UserMixin:
    def get_create_user(self, obj):
        # call get_user_data() here

    def get_update_user(self, obj):
        # call get_user_data() here

```
This way, it should be the case that the serializer classes that inherit from this don't even need to define `get_create_user` or get_update_user` explicitly, and you should be able to use `get_user_data` directly in `services/model_year_report.py` without needing to instantiate the class.

Please let me know if any issues come up or if you have any questions!

tim738745 (changes requested):

tim738745 (approved):

PR #2370, opened 2024-11-29
task: 2363 Flagging Duplicate VINs in BCeID Spreadsheet Uploads to Prevent Error 31 (duplicate VINs)

emi-hi:
spreadsheets with duplicate vins are rejected with error message during submission

### DIFF
--- a/backend/api/services/sales_spreadsheet.py
+++ b/backend/api/services/sales_spreadsheet.py
@@ -347,18 +347,26 @@ def validate_spreadsheet(data, user_organization=None, skip_authorization=False)
         )
 
     row = start_row
+    vin_dict = {}
+    duplicates = {}
 
     while row < min((MAX_READ_ROWS + start_row), sheet.nrows):
         row_contents = sheet.row(row)
         model_year = str(row_contents[0].value).strip()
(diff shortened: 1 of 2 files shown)


backend/api/services/sales_spreadsheet.py
```diff
@@ -347,18 +347,26 @@ def validate_spreadsheet(data, user_organization=None, skip_authorization=False)
...
         row_contains_content = False
         if len(model_year) > 0 or len(make) > 0 or len(model_name) > 0 or \
                 len(date) > 0:
             row_contains_content = True
+        if row_contains_content:
+            if vin in vin_dict:
+                vin_dict[vin].append(row + 1)
+            else:
+                vin_dict[vin] = [row + 1]
+        duplicates = {vin: rows for vin, rows in vin_dict.items() if len(rows) > 1}
```
tim738745, line 369: Can we move this line outside of the while loop (e.g. to line 382)? I think we can do this after vin_dict has been fully constructed, and that way we won't have to iterate though partial vin_dicts for each row in the spreadsheet

tim738745 (changes requested):

tim738745 (approved):

PR #2401, opened 2024-12-18
feat: 2373 - adds separate comment box for reject/request changes

emi-hi:


tim738745 (approved):

frontend/src/vehicles/components/VehicleDetailsPage.js
```diff
@@ -25,129 +25,132 @@ const VehicleDetailsPage = (props) => {
...
+    },
+  };
+
+  const getModalProps = (type) => ({
+    confirmLabel: modalConfig[type]?.confirmLabel || "",
+    buttonClass: modalConfig[type]?.buttonClass || "",
+    modalText: modalConfig[type]?.modalText || "",
+    handleSubmit: modalConfig[type]?.handleSubmit || (() => {}),
+  });
+  const modalProps = getModalProps(modalType);
```
tim738745, line 146: Since we're refactoring this modalProps thing, maybe we can make `modalProps` a state and set it using a `useEffect`, just to make it a bit more React-ish? so perhaps something like:
```
const [modalProps, setModalProps] = useState[{/*default modal props here*/}]
...
useEffect(() => {
  const newModalProps = getModalProps(modalType)
  setModalProps(newModalProps)
}, [modalType])
```

tim738745 (changes requested):

tim738745 (approved):

PR #2402, opened 2024-12-18
2339 - Fix tab/stage-name inconsistency during loading issue

rogerlcleung:
The model-year is now included in the query string when navigating between the compliance tabs so that the containers can get the model-year and determine the correct tab name before loading is finished.


frontend/src/compliance/AssessmentContainer.js
```diff
@@ -1,23 +1,23 @@
...
+import getComplianceObligationDetails from "../app/utilities/getComplianceObligationDetails";
+import ROUTES_SUPPLEMENTARY from "../app/routes/SupplementaryReport";
+import { getNewBalancesStructure, getNewDeficitsStructure } from "../app/utilities/getNewStructures";
+import getTotalReductionBig from "../app/utilities/getTotalReductionBig";
+import getClassAReductionBig from "../app/utilities/getClassAReductionBig";
+import getUnspecifiedClassReductionBig from "../app/utilities/getUnspecifiedClassReductionBig";
+import { convertBalances, convertCarryOverDeficits } from "../app/utilities/convertToBig";
+import getSnapshottedComplianceRatioReductions from "../app/utilities/getSnapshottedReductions";
 
 const AssessmentContainer = (props) => {
```
tim738745, line 22: Just wondering if there is a reason why we don't pass the `year` to `AssessmentContainer` via `ComplianceReportTabs` like we do for the other containers?

tim738745 (changes requested):

frontend/src/compliance/AssessmentContainer.js
rogerlcleung: Currently (before the changes), when clicking on the Summary page or the Assessment page, the tabs will disappear during loading. Therefore, the original "inconsistent tab name during loading" issue can only be observed when navigating between the first three tabs, but not the Summary page and the Assessment page. After discussing with Lalita and Shaylene, we agreed that the inconsistency between this appearing/disappearing of the tabs should not occur in the Summary page. Fixing this was included in this PR. The layout of the Assessment page is different as there are other things on top of the tabs. In this case, keeping the tabs disappeared during loading might make sense. We agreed to leave this behaviour as in for the Assessment page as this would need more thoughts on the correct behaviour.

Passing the "year" to the containers is only required for the container to determine the displayed tab names during loading. Since the tabs will not be appeared when loading the Assessment page, I decided to leave the Assessment page untouched. If later we come up with a new behaviour for loading up this page, we can modify the code to fit the needs at that time.

frontend/src/app/routes/Compliance.js
```diff
@@ -36,3 +36,8 @@ const COMPLIANCE = {
 }
 
 export default COMPLIANCE
+
+export const insertIdAndYear = (route, id, modelYear) => {
```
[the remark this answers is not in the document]
rogerlcleung: Done.

tim738745 (approved):

Issue #2398, opened 2024-12-18
ZEVA - Improve Default Row Display Behavior in Credit Application Table


shayjeff:
**Describe the task**
Fix the default row display behavior in the credit application table so it adjusts to the actual size of the dataset. Currently, it always defaults to 100 rows, even when there are only a few rows. This wastes space and is not user-friendly. The table should instead default to the next closest option from the available row numbers: 5, 10, 15, 20, 25, 50, 100.

For example:
- If there are 5 rows of data, it should default to 5.
- If there are 18 rows, it should default to 20 (the next closest option).
- If there are 2000 rows, it should default to 100 rows per page, and the user can use pagination to move to the next set of rows.

**Purpose**
This update will make the table cleaner and more user-friendly for datasets of any size. For smaller datasets, it eliminates unnecessary empty space. 

**Acceptance Criteria**
- [ ] The table defaults to the next closest option based on the number of rows in the dataset (available options: 5, 10, 15, 20, 25, 50, 100).
- [ ] If the dataset has fewer rows than the smallest option (5), it defaults to 5.
- [ ] If the dataset falls between two options (e.g., 18 rows), it rounds up to the next closest choice (e.g., 20).
- [ ] The dropdown for selecting rows (5, 10, 15, etc.) continues to function as expected.

**Additional context**

shayjeff:
Tested and approved. All criteria met with no issues found.

PR #2411, opened 2025-01-06
2398: Limiting Displayed Rows in Credit Application

rogerlcleung:
Refer to comments in #2398 for the alternative solution being implemented.

### DIFF
--- a/frontend/src/credits/components/VINListTable.js
+++ b/frontend/src/credits/components/VINListTable.js
@@ -323,6 +323,7 @@ const VINListTable = (props) => {
       page={page - 1}
       pages={calculateNumberOfPages(itemsCount, pageSize)}
       pageSize={pageSize}
+      minRows={1}
       sorted={sorts}
       filtered={filters}
       onPageChange={(pageIndex) => {


frontend/src/credits/components/VINListTable.js
```diff
@@ -323,6 +323,7 @@ const VINListTable = (props) => {
       page={page - 1}
       pages={calculateNumberOfPages(itemsCount, pageSize)}
       pageSize={pageSize}
+      minRows={1}
```
tim738745, line 326: I think this just makes it so that there are no padding rows (https://github.com/TanStack/table/blob/v6/src/index.js#L129). The card asks for there to be some padding rows (e.g. 48 total records, page size of 20 => 2 padding rows on the last (3rd) page, since, of the page size options, 10 is the "ceiling" of 8 (the number of records on the last page)). In order to do what the card specifies, I think we have to calculate `minRows` by taking into account `items.length`, `pageSizeOptions`, and `pageSize`. However, if you don't think it's worth it to do this given that the intent behind the card is to just make the table easier to read, please connect with Shaylene to discuss with her if you haven't already. Please let me know what you decide! Thanks!

tim738745: Oh never mind, I just saw your comment on the alternative solution!

PR #2414, opened 2025-01-09
feat: 2409 -  disables save button if forecast isn't filled, adds tooltip

emi-hi:
A flag is triggered if the Forecasting Table is not completed. **Disables save button** 

Suppliers cannot submit the Model Year report if this flag is active. **Cannot save the page if totals arent filled**

The flag message clearly states that the Forecasting Table must be filled out. **Displays tool tip**

The functionality applies to the Consumer ZEV Sales page for MY2023 and the ZEVs Supplied and Registered page for MY2024 and beyond.


jdtoombs (approved):
Looks good!

PR #2429, opened 2025-01-22
2415 - Function for Resetting the Database

rogerlcleung:
The function for resetting the database is added as an Django admin command.

To see the help of the function:
```
python manage.py help reset_database
```

To run the function:
```
python manage.py reset_database
```


backend/api/management/commands/reset_database.py
```diff
@@ -0,0 +1,223 @@
...
+            action = "store_true",
+            help = "also remove all user profiles without an organization."
+        )
+        parser.add_argument(
+            "--non-gov",
+            action = "store_true",
+            help = "also remove all vehicle suppliers (non-government organizations) and non-government users."
+        )
+
+    def handle(self, *args, **options):
```
tim738745, line 140: Can we add a check here for the environment so that if the environment = production, then the command doesn't run? Just to be safe! You can use settings.ENV_NAME (found in the settings.py file)

backend/api/models/__init__.py
```diff
@@ -15,7 +15,6 @@
 from . import vehicle_class
 from . import fixture_migration
 from . import vehicle_attachment
-from . import account_balance
```
tim738745, line 18: When deleting a model or making changes to a model, we should run `makemigrations` so that our changes are reflected in the database: https://realpython.com/django-migrations-a-primer/#creating-migrations

tim738745 (changes requested):

tim738745 (approved):

Issue #2329, opened 2024-10-23
ZEVA - update credit transfer unit tests


tim738745:
**Describe the task**
The api/fixture/test/0003_add_bceid_users.py file should be updated so that each test user is assigned a distinct organization; currently, 2 distinct users can be assigned the same organization and if that happens, some credit transfer tests will fail.

**Purpose**
Makes it so that our unit tests don't fail intermittently.

**Acceptance Criteria**
- [ ] The credit transfer unit tests pass.

**Developer Checklist**
(1) Update the code in the file mentioned above (0003_add_bceid_users.py) in the manner specified above.
(2) Update the `test_list_transfer` and `test_transfer_pass` tests in test_credit_transfers.py so that they don't use `transaction.on_commit()` (that is there because, earlier, that was Tim's suggested solution to the intermittent failure of these tests, but it actually makes it so that the assertions don't run because the transactions are never actually committed!).

PR #2438, opened 2025-01-29
ZEVA-2329:  Credit transfer unit tests

jdtoombs:
Unit tests for credit transfer on the backend. Still one problematic test that I need to investigate further to ensure it is not an underlying issue in the present application.


tim738745:
Hi @jdtoombs, this issue may help: https://github.com/bcgov/zeva/issues/2329. Each time the tests run, users get randomly assigned an organization, so 2 distinct users may belong to the same organization. Therefore, in this test file, org1 may equal org2, and so some tests may fail intermittently because they assume that the orgs are distinct.

jdtoombs:
> Hi @jdtoombs, this issue may help: #2329. Each time the tests run, users get randomly assigned an organization, so 2 distinct users may belong to the same organization. Therefore, in this test file, org1 may equal org2, and so some tests may fail intermittently because they assume that the orgs are distinct. So you could modify the `run` method in `api/fixtures/test/0003_add_bceid_users`:
> 
> ```
> @transaction.atomic
>     def run(self):
> 
>         organizations = list(Organization.objects.filter(
>             is_government=False,
>             is_active=True
>         ))
> 
>         users_added = 0
> 
>         for user in self.list_of_users:
>             organization = random.choice(organizations)
> 
>             (_, created) = UserProfile.objects.get_or_create(
>                 first_name=user.get("first_name"),
>                 last_name=user.get("last_name"),
>                 organization=organization,
>                 username=user.get("username"),
>                 keycloak_user_id=user.get("keycloak_user_id"),
>                 defaults={
>                     "is_active": True,
>                     "display_name": user.get("display_name")
>                 }
>             )
> 
>             organizations.remove(organization)
> 
>             if created:
>                 users_added += 1
> 
>         print("Added {} BCEID Users.".format(users_added))
> ```

Perfect, thank you! Will take a look

tim738745 (approved):

PR #2451, opened 2025-02-10
feat: update organization sales/supplied average to display sales if less than 3 years of supplied

emi-hi:
-if less than 3 years of records exist for supplied, passes sales instead (dont just show 0)
-new @property to display whether its a sales or supplied average 
-updates frontend to display whether its sales or supplied

### DIFF
--- a/backend/api/models/model_year_report.py
+++ b/backend/api/models/model_year_report.py
@@ -170,9 +170,7 @@ class ModelYearReport(Auditable):
         ).first()
 
     def get_avg_sales(self):
-        avg_sales = self.organization.get_avg_ldv_sales(
-            year=self.model_year.name
-        )
+        avg_sales, _ = self.organization.get_avg_ldv_sales(year=self.model_year.name) or (None, None)
 
         # if this is empty that means we don't have enough ldv_sales to
         # get the average. our avg_sales at this point should be from the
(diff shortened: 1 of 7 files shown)


tim738745 (changes requested):
@emi-hi, I think there are some references to `get_avg_ldv_sales()` in the backend that need to be addressed as well (line 173 in models/model_year_report.py, as well as line 181 in viewsets/model_year_report.py).

Also, I was thinking we don't need to update the frontend `avgLdvSales` references; maybe, in the `OrganizationSerializer` (line 23 of serializers/organization.py), we could do:
```
class OrganizationSerializer(serializers.ModelSerializer):
  ...
  avg_ldv_sales = serializers.SerializerMethodField()
  supplied_or_sales = serializers.SerializerMethodField()
  ...
  def get_avg_ldv_sales(self, obj):
        avg_sales, _ =  obj.get_avg_ldv_sales()
        return avg_sales

  def get_supplied_or_sales(self, obj):
        _ , supplied_or_sales = obj.get_avg_ldv_sales()
        return supplied_or_sales
  ...
  class Meta:
        model = Organization
        fields = (... "avg_ldv_sales", "supplied_or_sales",...)
```
And on the frontend, in the VehicleSupplierDetailsPage, just have what you had before:
`<h4 className="d-inline">3 Year Average Vehicles {details.suppliedOrSales}: </h4>`

tim738745 (approved):

PR #2474, opened 2025-03-06
fix overflowing tooltip

jdtoombs:
Some tooltips were overflowing on a single line when it was over 500px, this ensures that it will break the word and flow down.


jdtoombs:
@shayjeff @Lalita-Lala - this pr should fix the issues!

tim738745 (approved):

Issue #2485, opened 2025-03-26
ZEVA -Disable "Delete" Button for Credit Applications on BCeID Side


shayjeff:
**Describe the task**
Remove or disable the ability to delete credit applications on the BCID side of the system.

**Purpose**
To prevent BCeID users from deleting credit applications, which can unintentionally impact internal records on the IDIR side.

**Acceptance Criteria**
- [x] The "Delete" button is removed or disabled for credit applications on the BCeID credit application tab.

**Additional context**
- Deleting rejected credit applications from the BCeID side currently removes internal records from IDIR
- Keeping records on the IDIR side is important for audit and validation tracking

<img width="946" alt="Image" src="https://github.com/user-attachments/assets/fb02c175-298a-47d9-b988-850e5c776e9c" />

danclancy85:
Tested

PR #2486, opened 2025-03-27
Task: Disable Credit Application Delete Button From BCEID Side #2485

JulianForeman:
Having some issues logging into the bceid accounts at the moment but I know this is high priority. So, this should be the fix for it but if possible someone should double check locally before merging it in.

- Disabling credit application delete button from BCEID side

### DIFF
--- a/frontend/src/credits/components/CreditRequestDetailsPage.js
+++ b/frontend/src/credits/components/CreditRequestDetailsPage.js
@@ -576,7 +576,7 @@ const CreditRequestDetailsPage = (props) => {
               {(submission.validationStatus === 'DRAFT' ||
                 submission.validationStatus === 'REJECTED') &&
                 typeof user.hasPermission === 'function' &&
-                user.hasPermission('EDIT_SALES') && (
+                user.hasPermission('EDIT_SALES') && user.isGovernment && (
                   <Button
                     buttonType="delete"
                     action={() => {


frontend/src/credits/components/CreditRequestDetailsPage.js
```diff
@@ -576,7 +576,7 @@ const CreditRequestDetailsPage = (props) => {
               {(submission.validationStatus === 'DRAFT' ||
                 submission.validationStatus === 'REJECTED') &&
                 typeof user.hasPermission === 'function' &&
-                user.hasPermission('EDIT_SALES') && (
+                user.hasPermission('EDIT_SALES') && user.isGovernment && (
```
tim738745, line 579: I haven't tested this, but it looks like this change would mean supplier users can't delete draft credit applications, which I don't think is correct? Perhaps chat with Shaylene about this? Thanks!

tim738745 (changes requested):

frontend/src/credits/components/CreditRequestDetailsPage.js
JulianForeman: Yeah, after testing it I realized we can just remove the or status is rejected check and this works as intended.

tim738745 (approved):

PR #2488, opened 2025-03-31
Task 2471 - Uploading Updated ZEV Flag Datasets from ICBC (October 2021 – June 2023)

emi-hi:


tim738745 (commented):
@emi-hi looks good to me! Just want to make sure I understand the process here: 

(1) The `compare_csvs` function was run 21 times - once for each ICBC file, and each time the file was compared with the same output file generated from the SQL in the first cell.

(2) Each vin in the collection of 109 additional vins comes from at least one of the comparison instances, not necessarily every one of the comparison instances. For example, if vin X **is not** in our current icbc registration data, it **is** in the 2022-01-31 file, and it **is not** in the 2022-08-31 file, then vin X would be a member of the 109 vins?

PR #2571, opened 2026-01-09
Task 2545 - ICBC Upload User Notification

dallascrichmond:
Before changes:
- When a user uploaded a large ICBC file the progress bar would immediately go from 0%-100% and then appear as if frozen.
- If the user refreshed or navigated away from the ICBC upload tab the progress bar would disappear and there would be no confirmation that the process was complete and the current date was not updated. This is because the upload process was still occurring in the background, which would eventually be reflected in the current dates once complete, but there was no persistence of the progress state.

After Changes:
- Now the progress bar polls the backend every 10 seconds (Was the easiest way that I could find that doesn't over complicate things) and gives the user a more accurate progress bar and details (Processing 1 out of 5 pages..)
- If the user refreshes the page or navigates away, local storage is used to see if the process is still ongoing, which then allows the progress bar to remain active in the ICBC upload tab until complete.
- I would recommend testing with a large ICBC file, but be aware that although it may appear as though the progress bar is "frozen" at something like 50%, each chunk of data has thousands of entries and it can sometimes take up to 5 minutes for a chunk or "page" to be processed.

### DIFF
--- a/backend/api/viewsets/icbc_verification.py
+++ b/backend/api/viewsets/icbc_verification.py
@@ -3,193 +3,105 @@ import os
 import threading
 
 from django.http import HttpResponse
-from django.db import connection
-from django.utils import timezone
 from rest_framework import viewsets, status
 from rest_framework.decorators import action
 from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 
-from api.services.icbc_upload import ingest_icbc_spreadsheet, get_upload_progress, set_upload_progress
-from api.services.minio import get_minio_object, minio_remove_object
+from api.services.icbc_upload import (
+    get_upload_progress,
+    set_upload_progress,
+    process_upload,
+)
 from api.models.icbc_upload_date import IcbcUploadDate
-from api.models.icbc_upload_progress import IcbcUploadProgress
 from api.serializers.icbc_upload_date import IcbcUploadDateSerializer
-from api.serializers.icbc_upload_progress import IcbcUploadProgressSerializer
+
 
 class IcbcVerificationViewSet(viewsets.GenericViewSet):
     permission_classes = [AllowAny]
-    http_method_names = ['get', 'post']
+    http_method_names = ["get", "post"]
 
-    serializer_classes = {
-        'default': IcbcUploadDateSerializer
-    }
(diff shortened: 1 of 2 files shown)


tim738745 (changes requested):

backend/api/migrations/0013_auto_20260115_1107.py
```diff
@@ -0,0 +1,94 @@
...
+                ('results', models.JSONField(blank=True, null=True)),
+                ('created_at', models.DateTimeField(auto_now_add=True)),
+                ('updated_at', models.DateTimeField(auto_now=True)),
+            ],
+            options={
+                'db_table': 'icbc_upload_progress',
+                'ordering': ['-created_at'],
+            },
+        ),
+        migrations.AlterField(
```
tim738745: Can we remove the `AlterField`s in this file? I think these are generated because we updated the django version a while back, and in that version, foreign key fields use `bigint` instead of `int`. We should apply these migrations sooner or later, but I think I'd prefer to keep it separate from this PR.

backend/api/viewsets/icbc_verification.py
```diff
@@ -56,56 +122,128 @@ def upload(self, request):
...
+                )
+
+                if done[0]:
+                    # We remove the previous file from minio but keep the 
+                    # latest one so we can use it for compare on next upload
+                    minio_remove_object(previous_filename)
+                    print('Done processing')
+                    
+                    from django.db import connection
+                    with connection.cursor() as cursor:
```
tim738745: Here as well, is there a reason we prefer raw sql over django's queryset API?

backend/api/models/icbc_upload_progress.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+    error = models.TextField(
+        null=True,
+        blank=True
+    )
+    results = models.JSONField(
+        null=True,
+        blank=True
+    )
+    created_at = models.DateTimeField(
```
tim738745: We have an `Auditable` model we can inherit from (`backend/auditable/models.py`), so we don't have to manually define create and update timestamps for each new model!

